PR #52, opened 2023-09-07
Add api key from extension & Workspace from extension

pavel-pythagora:


LeonOstrez (changes requested):
for consistency reasons we should add `PYTHAGORA_API_KEY` to .env.example (.env) and then just have `api_key = os.getenv("PYTHAGORA_API_KEY") or os.getenv("OPENAI_API_KEY")`

pavel-pythagora:
@LeonOstrez Please specify what you mean. Do you want me to assign api key to enviroment variable instead of regular variable? And I don't understand why do i need to use PYTHAGORA_API_KEY, as if we use OPENAI_API_KEY

pilot/const/config.py
```diff
@@ -0,0 +1,2 @@
+API_KEY = ''
```
nalbion, line 1: `OPENAI_API_KEY` and `AZURE_API_KEY` are specified in pilot/.env

pilot/main.py
```diff
@@ -74,6 +75,11 @@ def local_print(*args, **kwargs):
     args = init()
 
     builtins.print, ipc_client_instance = get_custom_print(args)
+    if '--api-key' in args:
+        const.config.API_KEY = args['--api-key']
+
+    if '--root-folder' in args:
```
nalbion, line 81: `workspace` can now be specified as a CLI arg

https://github.com/Pythagora-io/gpt-pilot/blob/main/pilot/utils/arguments.py#L31

nalbion (changes requested):

pilot/main.py
pavel-pythagora: Yes, but i create this branch from `ipc` branch. And `ipc` branch does not have this functionality.

nalbion: what's the `ipc` branch for?

pavel-pythagora: For ipc. We use ipc for connection beetween extension and cli

nalbion: what extensions?

nalbion: I think 've added `workspace` with the same idea in mind. An IDE plugin would be able to run GPT Pilot with `workspace=/path/to/project` and the `user_apps` table can be used to load an App config record by `workspace` and `user_id`.

nalbion: @pavel-pythagora I'd be interested to get your thoughts on https://github.com/Pythagora-io/gpt-pilot/issues/89

PR #54, opened 2023-09-07
get email from ~/.gitconfig

nalbion:
If users have their git user set up with an email address, this can be used as the default email.


pilot/utils/arguments.py
```diff
@@ -36,12 +39,10 @@ def get_arguments():
         arguments['app_id'] = str(uuid.uuid4())
 
     if 'user_id' not in arguments:
-        arguments['user_id'] = str(uuid.uuid4())
+        arguments['user_id'] = getpass.getuser()
```
LeonOstrez, line 42: since https://github.com/Pythagora-io/gpt-pilot/pull/57 is merged we dont need this, right?

LeonOstrez (changes requested):
can you please resolve conflicts and i can merge it then. Thank you!

pilot/utils/arguments.py
nalbion: yeah, it must have leaked in from the other branch

nalbion:
Ready for merge

PR #69, opened 2023-09-08
feat(llm_connection): add openrouter api endpoint

mrgoonie:
I added [OpenRouter API](https://openrouter.ai/) endpoint so we can use other LLM models (such as Google Bard, Claude,...)


pilot-env/pyvenv.cfg
LeonOstrez, line 1: how come you commited whole pilot-env folder?

pilot/.env.example
```diff
@@ -1,13 +1,24 @@
-#OPENAI or AZURE
-ENDPOINT=OPENAI
+# OPENAI or AZURE or OPENROUTER
+ENDPOINT=OPENROUTER
```
LeonOstrez, line 2: let's try to keep all defaults same as before. So in this case ENDPOINT=OPENAI

pilot/.env.example
```diff
@@ -1,13 +1,24 @@
...
+
+OPENAI_ENDPOINT=
 OPENAI_API_KEY=
+
 AZURE_API_KEY=
 AZURE_ENDPOINT=
-#In case of Azure endpoint, change this to your deployed model name
-MODEL_NAME=gpt-4
```
LeonOstrez, line 7: same as above, lets keep default to gpt-4

pilot/.env.example
```diff
@@ -1,13 +1,24 @@
...
+
+OPENROUTER_API_KEY=
+OPENROUTER_ENDPOINT=https://openrouter.ai/api/v1/chat/completions
+
+# In case of Azure/OpenRouter endpoint, change this to your deployed model name
+MODEL_NAME=openai/gpt-3.5-turbo-16k
 MAX_TOKENS=8192
+
+# Database
+DATABASE_TYPE=postgres
```
LeonOstrez, line 18: same as above lets keep db sqlite

pilot/.env.example
mrgoonie: Done

pilot-env/pyvenv.cfg
mrgoonie: Added to `.gitignore` already

LeonOstrez (changes requested):
can you please remove whole pilot-env folder from pull request and then i can merge it. Thanks for making PR!

mrgoonie:
> I suppose support for Code Llama would be out of scope for this PR...

You mean this one?
image

OpenRouter supports many LLMs, you can check out at their website: https://openrouter.ai/docs

Issue #80, opened 2023-09-11
execute_step(matching_step, current_step) is confusing


nalbion:
I think that `execute_step(matching_step, current_step)` should be renamed `should_execute_step(arg_step, current_step)`.

Also, this new `test_no_step_arg()` that I've written fails - am I misunderstanding the intention?

```python
class TestExecuteStep:
    def test_no_step_arg(self):
        assert execute_step(None, 'project_description') is True
        assert execute_step(None, 'architecture') is True
        assert execute_step(None, 'coding') is True

    def test_skip_step(self):
        assert execute_step('architecture', 'project_description') is False
        assert execute_step('architecture', 'architecture') is True
        assert execute_step('architecture', 'coding') is True

    def test_unknown_step(self):
        assert execute_step('architecture', 'unknown') is False
        assert execute_step('unknown', 'project_description') is False
        assert execute_step('unknown', None) is False
        assert execute_step(None, None) is False
```

PR #81, opened 2023-09-11
Refactored `execute_step()` as per #80

nalbion:
Refactored `execute_step()` as per #80

### DIFF
--- a/pilot/helpers/agents/Architect.py
+++ b/pilot/helpers/agents/Architect.py
@@ -4,11 +4,13 @@ import json
 from termcolor import colored
 from const.function_calls import ARCHITECTURE
 
-from utils.utils import execute_step, find_role_from_step, generate_app_data
+from utils.utils import should_execute_step, find_role_from_step, generate_app_data
 from database.database import save_progress, get_progress_steps
 from logger.logger import logger
 from prompts.prompts import get_additional_info_from_user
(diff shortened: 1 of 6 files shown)


pilot/helpers/agents/Architect.py
```diff
@@ -4,32 +4,34 @@
...
+        step = get_progress_steps(self.project.args['app_id'], ARCHITECTURE_STEP)
+        if step and not should_execute_step(self.project.args['step'], ARCHITECTURE_STEP):
             step_already_finished(self.project.args, step)
             return step['architecture']
 
         # ARCHITECTURE
         print(colored(f"Planning project architecture...\n", "green", attrs=['bold']))
         logger.info(f"Planning project architecture...")
 
+        self.convo_architecture = AgentConvo(self)
```
LeonOstrez, line 34: why did you move setting of convo_architecture here?

In this case if we load this step from DB, in later steps convo_architecture will be undefined. I think with current implementation it might not matter but would be good to have it for future

nalbion: Because I'm used to Sonarqube complaining that variables are declared too far from where they are first used.

Following YAGNI, `convo_architecture` should probably be a local variable _until_ we do actually have a need for it to be an instance variable.

Issue #91, opened 2023-09-13
Initial prompt routing


nalbion:
As a user 
I want to start GPT Pilot with various types of initial prompts
So that I can build a new app, modify an existing app or debug an issue.

nalbion:
After studying the documentation and source for [LangChain Agents](https://docs.langchain.com/docs/components/agents) and Microsoft's [AutoGen](https://github.com/microsoft/autogen) I've come up with an idea for how we could make some architectural changes to allow a lot more flexibility in the future.

![image](https://github.com/Pythagora-io/gpt-pilot/assets/1866161/601a85c6-8753-43b9-b791-8fd693b6c47c)

nalbion:
The objective being that we end up with an extensible framework which integrates with existing tools and agents:

![image](https://github.com/Pythagora-io/gpt-pilot/assets/1866161/2e19a407-235e-4f89-ae6c-f25072a940e2)

mnieber:
It would be interesting to consider how such an architecture could support the existing use-case (gpt pilot with its current agents) without really changing anything, so that the "open" case (where you can add any agents to the system) is a generalization of the "closed" case (gpt pilot with a fixed set of agents).

For example, TechLead could describe the other agents that it depends upon. It can declare that it needs to work together with a Developer agent, and offer a description of what a Developer agent should be capable of. In that case, TechLead can still send messages to Developer (*), but the router will intercept those messages and find out if there is any available agent that can take this Developer role. This requires that every agent is able to describe itself, for example using the Abilities that AutoGPT uses.

(*) in this case, TechLead only has an abstract notion of "Developer". It doesn't know which concrete Developer agent it is communicating with.

nalbion:
GitHub CoPilot now lets you address agents directly using `@workspace` or `@vscode` in addition to `/slash` commands which the agents support

wayland:
@nalbion Love your work!  

@mnieber Interesting idea (where they declare what agents they want to work with).  Could tie in well with my comments at https://github.com/Pythagora-io/gpt-pilot/issues/182#issuecomment-1836435382 .

Issue #89, opened 2023-09-12
Implement "Agent Protocol"


nalbion:
GPT Pilot _may_ be eligible for the [AutoGPT Arena Hacks](https://lablab.ai/event/autogpt-arena-hacks) if it implements the [Agent Protocol](https://agentprotocol.ai/)

There are [client SDKs](https://agentprotocol.ai/clients) (coming), for now [Python](https://agentprotocol.ai/clients/python) seems to be more supported.

There are [other reasons](https://agentprotocol.ai/incentives) why it would be good to adopt a common interface.

A `Task` denotes one specific goal for the agent, it can be specific like:

> Create a file named `hello.txt` and write `World` to it.

or very broad as:

> Book a flight from Berlin to New York next week, optimize for price and duration.

- POST /agent/tasks - for creating tasks
```python
{
  task_id: str, 
  input: str,
  additional_input: {},
  steps: [, ...],
  artifacts: [{
    artifact_id: str,
    file_name: str,
    relative_path: str,
  }, ...]
}  
```

- POST /agent/tasks/{id}/steps - for triggering next step for the task
```
{
    task_id: str,
    step_id: str,
    input: str,
    additional_input: str,
    name: str,
    status: 'created' | 'completed',
    output: '',
    additional_output: {},
    artifacts: [{ as below }, ...],
    is_last: boolean,
}
```

nalbion:
This API also fits well with my `StoryManager` concept illustrated in #83

nalbion:
There's only a partial alignment with the current naming:

<table>
<tr><th>GPT Pilot</th><th>Agent Protocol</th>
<tr><th>db.model.UserStories</th><th>Task</th></tr>
<tr><td>project.user_stories</td><td></td></tr>
<tr><td>{app_id}-{i} or {app_id}#{jira/github issue}</td><td>task_id</td></tr>
<tr><td>story title</td><td>input</td></tr>
<tr><td>app_id, user_id, sprint</td><td>additional_input</td></tr>
<tr><td></td><td>steps</td></tr>
<tr><td></td><td><b>artifacts</b></td></tr>


<tr><th>db.model.DevelopmentSteps</th><th>Step</th></tr>
<tr><td>(just used for replaying?)</td><td></td></tr>
<tr><td>as above</td><td>task_id</td></tr>
<tr><td>id</td><td>step_id</td></tr>
<tr><td></td><td>is_last</td></tr>
[...]

PR #94, opened 2023-09-13
Receptionist routes to the best agent for initial user input

nalbion:
Implementation for #91 

> Write the word 'Washington' to a .txt file

Can go straight to `CodeMonkey` so that we can be benchmarked against ChatGPT agents.

> A simple chat app with real time communication

Needs to go to the `ProductOwner`.

# Introducing the Receptionist into the conversation
NOTE - This isn't wired up yet. To pass the benchmarks, The `Receptionist` needs to take the first user input prompt.

The problem is that the call flow, by default goes:

- PO: What is the project name?
- User: My App
- PO: Describe your app in as many details as possible.
- User: A simple chat app with real time communication

It would need to be changed to:

- PO: Describe your app in as many details as possible.
- User: A simple chat app with real time communication
( PO asks Receptionist "do you think they want to speak to me or the Code Monkey?" )
- PO: What is the project name?
- User: My App

Or:

- Reception: Hi, how can I help you?
- User: I want to build a web app
- PO: What is the project name?
- User: My App
- PO: Describe your app in as many details as possible.
- User: A simple chat app with real time communication

Or:

- Reception: Would you like to build a new project or 
- User: I want to build a web app
- PO: What is the project name?
- User: My App
- PO: Describe your app in as many details as possible.
- User: A simple chat app with real time communication

# Agent Protocol
I'm also thinking how the Agent Protocol (#89) would work. My original thinking was:

An IDE "New Project" wizard could have a UI which asked for the the following and sent them as sequential `input`s and possibly `additional_input` fields, but I'm not a big fan of using the `additional_input` as it's something that the client (IDE) needs to know about, but it _could_ be a good mechanism to getting past the Receptionist.

- workspace path: /src/chat-app`
- project name: chat app
- description: A simple chat app with real time communication

```
POST /agent/tasks
{
   "input": "chat app"                      // args.name
   "additional_input": {
     "workspace": "/src/chat-app"   // args.workspace
     "task_type": "Web App"
   }
}
```

Creates an `App` and responds with `app_id` as `task_id`
```> { "task_id": "chat_app" }```

Is it reasonable that the client would know (how) to send the `Description` from the New Project wizard UI?

It's possible that the client just starts off with an empty text box like ChatGPT, but at this stage I'm not sure what the client would be sending in the initial Task post.

We still need to cater for the standard use which seems to be:

- user begins a new conversation without any guidance or prompting from our app or LLM, by `POST /agent/tasks`
- user sends further inputs to `POST /agent/tasks/{task_id}/steps`

The catch is that we're not able to present them with a prompt from our app/LLM until they've POSTed the first `steps` input.

```
POST /agent/tasks/chat_app/steps
{
   "input": "A simple chat app with real time communication"
}
```

### DIFF
--- a/pilot/const/function_calls.py
+++ b/pilot/const/function_calls.py
@@ -112,6 +112,29 @@ COMMANDS_TO_RUN = {
     },
 }
 
+ROUTE_INITIAL_INPUT = {
+    'definitions': [
+        {
+            'name': 'route_initial_input',
+            'description': 'Routes the user to the agent best suited to handle a call.',
+            'parameters': {
+                'type': 'object',
+                'properties': {
+                    'agent': {
+                        'type': 'string',
+                        'description': 'Name of the agent that is best suited to handle a call.',
+                        'enum': ['ProductOwner', 'Architect', 'TechLead', 'Developer', 'CodeMonkey'],
+                    },
+                },
+                'required': ['agent'],
+            },
+        },
+    ],
+    'functions': {
+        'route_initial_input': lambda agent: agent
+    },
+}
+
 DEV_TASKS_BREAKDOWN = {
     'definitions': [
         {
--- a/pilot/helpers/AgentConvo.py
+++ b/pilot/helpers/AgentConvo.py
@@ -42,6 +42,7 @@ class AgentConvo:
         # craft message
         self.construct_and_add_message_from_prompt(prompt_path, prompt_data)
 
+        # TODO: should all of this move into the Developer class?
         # check if we already have the LLM response saved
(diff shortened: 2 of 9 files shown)


nalbion:
I noticed that `llm_connection.get_prompt` adds `no_micro_services` and `single_question` to the prompt, but it's not needed.

nalbion:
Actually, I like the idea of opening with the project description and allow the ProductOwner to suggest a name, this could even become a longer conversation ("I'm not sure" -> "how about these suggestions: ..." -> "No, it should..."):

User: A simple chat app with real time communication
PO: What is the project name? (press Enter for "Chat app")

Issue #99, opened 2023-09-20
Remove function calling


nalbion:
Remove function calling so that we can properly test other LLMs

nalbion:
Will need to revert back from `fabulous` to `colored` so that the code can be ran and debugged on Windows

https://github.com/jart/fabulous/issues/21

nalbion:
This is a model trained for function calling, based on Llama 2 - https://github.com/MeetKai/functionary

nalbion:
@LeonOstrez

In Developer.set_up_environment() there is code (which is currently unreachable, due to a `return` at line 312) which uses function calling for one of the few non-trivial use cases. It calls `execute_command_and_check_cli_response()` and introduces `send_convo` which creates a bit of a mess. 

...I'd like to refactor `AgentConvo.postprecess_response()` into `function_calling.process_json_response()` to make it clear that it's the complement to `function_calling.add_function_calls_to_request()`.

I'm going to refactor the part of Developer that sends `development/env_setup/install_next_technology.prompt` into a separate method and align it more with `Developer.implement_step()`

PR #102, opened 2023-09-22
Remove function calling

nalbion:
See also notes in #99

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ https://github.com/Pythagora-io/gpt-pilot/assets/10895136/0495631b-511e-451b-93d
 # 🔌 Requirements
 
 
-- **Python**
+- **Python >= 3.11**
 - **PostgreSQL** (optional, projects default is SQLite)
    - DB is needed for multiple reasons like continuing app development if you had to stop at any point or app crashed, going back to specific step so you can change some later steps in development, easier debugging, for future we will add functionality to update project (change some things in existing project or add new features to the project and so on)...
 
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ https://github.com/Pythagora-io/gpt-pilot/assets/10895136/0495631b-511e-451b-93d
 # 🔌 Requirements
 
 
-- **Python >= 3.8**
+- **Python >= 3.10**
 - **PostgreSQL** (optional, projects default is SQLite)
    - DB is needed for multiple reasons like continuing app development if you had to stop at any point or app crashed, going back to specific step so you can change some later steps in development, easier debugging, for future we will add functionality to update project (change some things in existing project or add new features to the project and so on)...
 
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ After you have Python and PostgreSQL installed, follow these steps:
 1. `git clone https://github.com/Pythagora-io/gpt-pilot.git` (clone the repo)
 2. `cd gpt-pilot`
 3. `python -m venv pilot-env` (create a virtual environment)
-4. `source pilot-env/bin/activate` (activate the virtual environment)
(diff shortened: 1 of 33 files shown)


README.md
```diff
@@ -51,7 +51,7 @@ https://github.com/Pythagora-io/gpt-pilot/assets/10895136/0495631b-511e-451b-93d
 # 🔌 Requirements
 
 
-- **Python**
+- **Python >= 3.10**
```
LeonOstrez, line 54: why python 3.10? I am using 3.9.6 on my current macbook and it works fine

pilot/helpers/agents/Developer.py
```diff
@@ -324,48 +324,62 @@ def set_up_environment(self):
...
-                        },
-                    }],
-                    'functions': {
-                        'execute_command': execute_command_and_check_cli_response
-                    },
-                    'send_convo': True
-                })
+            llm_response = self.install_technology(technology)
 
+            # TODO: I don't think llm_response would ever be 'DONE'?
```
LeonOstrez, line 329: you are right, once it is done it will return 'INSTALLED' which is defined in 'development/env_setup/install_next_technology.prompt'. This should be updated and maybe best way would be to put it in const/llm.py with other llm constants

LeonOstrez: but also have in mind that this feature is disabled at the moment (return on line 312) so it is not important to fix it now

pilot/helpers/cli.py
```diff
@@ -252,12 +258,15 @@ def execute_command_and_check_cli_response(command, timeout, convo):
 
     Returns:
         tuple: A tuple containing the CLI response and the agent's response.
+            - cli_response (str): The command output.
+            - llm_response (str): 'DONE' or 'NEEDS_DEBUGGING'
     """
-    cli_response, response = execute_command(convo.agent.project, command, timeout)
-    if response is None:
-        response = convo.send_message('dev_ops/ran_command.prompt',
+    # TODO: Prompt mentions `command` could be `INSTALLED` or `NOT_INSTALLED`, where is this handled?
```
LeonOstrez, line 264: for devops agent while installing tech dependencies response will be `INSTALLED` or `NOT_INSTALLED` but for developer agent, or in other words commands executed while coding eg. npm run start, it can return `DONE` or `NEEDS_DEBUGGING`.

pilot/utils/function_calling.py
```diff
@@ -0,0 +1,219 @@
...
+    definitions: list[FunctionType]
+    functions: dict[str, Callable]
+
+
+def add_function_calls_to_request(gpt_data, function_calls: FunctionCallSet | None):
+    if function_calls is None:
+        return
+
+    model: str = gpt_data['model']
+    is_llama = 'llama' in model or 'anthropic' in model
```
LeonOstrez, line 41: why is it different for llama compared to all other llms, or do we have to add implementation for each llm individually?

pilot/utils/function_calling.py
```diff
@@ -0,0 +1,219 @@
...
+
+    # if model == 'gpt-4':
```
LeonOstrez, line 43: why is this commented part here?

README.md
nalbion: I ran into an (minor) issue with typing and figured that Python 3.10 came out in 2021 so it would not be likely to be used anymore.

```
utils/function_calling.py:5: in <module>
    JsonType = str | int | float | bool | None | list["JsonType"] | dict[str, "JsonType"]
E   TypeError: unsupported operand type(s) for |: 'type' and 'type'
```

pilot/utils/function_calling.py
nalbion: I need to rename this to `is_instruct`. Wrapping in `[INST]` seemed to have a negative affect for some models.

nalbion:
I still need to fix the `colored` vs `fabulous` issue

PR #103, opened 2023-09-25
Project generators & GitHub template repos

nalbion:
Instead of relying on the AI to generate _all_ code for a new project, lean on the project generators & template repos relevant to each framework - eg:

- create-react-app etc
- flutter create
- gradle init
- spring init

Note: The `template_repo_embeddings.pkl` is quite large - maybe we should save to a public Vector DB such as Pinecone or Chroma

See also Pythagora-io/gpt-pilot-backend#2


nalbion:
This would be better with #63

nalbion:
https://blog.logrocket.com/using-supabases-vector-database-postgresql/

PR #125, opened 2023-10-03
fixed unit tests

nalbion:
Note that there are still some tests failing, and they are likely to have an impact on users.

https://github.com/Pythagora-io/gpt-pilot/actions

They were passing prior to this commit: https://github.com/Pythagora-io/gpt-pilot/commit/6c571f0946a0498ee9b623455034177ee7a4c589

I'm not sure would we want to replace `~` with `''` in `get_full_file_path()`

```python
        file_path = file_path.replace('~', '')
        file_name = file_name.replace('~', '')
```

We probably should have some safe-guards around the LLM reading/writing arbitrary files outside of the workspace folder, but it could be helpful if the the system was able to edit `~/.gitconfig`, `/etc/hosts` etc

Also, these lines are breaking some valid cases in `test_save_file()`

> {'name': 'Dockerfile', 'path': None, 'saved_to': '/temp/gpt-pilot-test/Dockerfile'},

```python
        if '.' not in file_path and not file_path.endswith('/'):
            file_path += '/'
        if '.' not in file_name and not file_name.endswith('/'):
            file_name += '/'
```

> {'name': '/etc/hosts', 'path': None, 'saved_to': '/etc/hosts'},

```python
        if '/' in file_path and not file_path.startswith('/'):
            file_path = '/' + file_path
        if '/' in file_name and not file_name.startswith('/'):
            file_name = '/' + file_name
```

### DIFF
--- a/pilot/utils/files.py
+++ b/pilot/utils/files.py
@@ -2,6 +2,7 @@ import os
 from pathlib import Path
 from database.database import save_user_app
 
+
 def get_parent_folder(folder_name):
     current_path = Path(os.path.abspath(__file__))  # get the path of the current script
 
@@ -11,7 +12,14 @@ def get_parent_folder(folder_name):
     return current_path.parent
 
 
-def setup_workspace(args):
+def setup_workspace(args) -> str:
+    """
+    Creates & returns the path to the project workspace.
+    Also creates a 'tests' folder inside the workspace.
+    :param args: may contain 'workspace' or 'root' keys
+    """
+    # `args['workspace']` can be used to work with an existing workspace at the specified path.
+    # `args['root']` is used by VS Code for (nearly) the same purpose, but `args['name']` is appended to it.
     workspace = args.get('workspace')
     if workspace:
         try:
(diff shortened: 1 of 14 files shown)


nalbion:
I'm not sure why `TestCodeMonkey` and `TestDeveloper` are failing in CI - they pass for me on Windows _and_ WSL Ubuntu

pilot/utils/files.py
```diff
@@ -23,7 +31,6 @@ def setup_workspace(args):
         return args['workspace']
 
     root = args.get('root') or get_parent_folder('pilot')
-    create_directory(root, 'workspace')
```
LeonOstrez, line 26: why is this removed?

nalbion: Because `os.makedirs` does this anyway

>def makedirs(name, mode=0o777, exist_ok=False):
>    """
>    Super-mkdir; create a leaf directory and all intermediate ones.  Works like
>    mkdir, except that any intermediate path segment (not just the rightmost)
>    will be created if it does not exist.

nalbion:
- added Python 3.12 to test matrix
- resolve #126  - Docker build requires 3.11 (for now)
- removed imports from `pilot/helpers/__init__.py` as it was breaking mocks on Python 3.9 & 3.10
- raise new `ApiKeyNotDefinedError`
- TODO: Treatment of paths outside of the project workspace - https://github.com/Pythagora-io/gpt-pilot/issues/129
- logger now logs at WARN and ERROR

PR #127, opened 2023-10-03
load app from the user given step

Dhanushkumar-S-G:
Resume app from the user given step if that step is within the the current app status, because when user specifying there are chances of skipping steps.


LeonOstrez:
This is great idea but I already merged https://github.com/Pythagora-io/gpt-pilot/pull/149 which was based on this PR.

Issue #114, opened 2023-09-28
[feature] Allow manual edits in development process


jonasdebeukelaer:
It would be nice to be able to allow manual edits during the debugging process, as sometimes it's quicker to fix a typo manually than to have the llm figure it out.

I think currently this is not possible because the state of the code is saved to the db, and so the code is reset after manual edits?

---
Tangentially, it would be great to be able to save the code change events as git commits. This would solve the above issue, and we could regain some of the nice features we get with git development flow.

nalbion:
I observed the other day that the system does pick up on manual changes

LeonOstrez:
you should be able to add your changes and gpt-pilot picks up on them when continuing development. If you found a case where that doesn't work as intended please open an issue and provide some more details on it. Thank you

jonasdebeukelaer:
Sorry not had much time to address this!

As linked by @nalbion, I think this was related to save/restore from db, as I was getting some errors when running which prob triggered this restore function. Looks like it's being sorted so happy to leave this one closed.

Issue #104, opened 2023-09-26
Project won't resume after interrupt


gmelfe:
When gpt-pilot execution gets interrupted (e.g. with CTRL+C), the project can not be resumed. 
When trying to resume the project with the app_id, I get prompted to describe the project again.
The project was already in the implementation stage (requirements were complete, architecture definition, etc.).

benrman:
Yeah I'm having the same issue.

nalbion:
@gmelfe @benrman does it work if you use `python main.py workspace=/src/my_project`?

benrman:
I am pretty sure that's what I did.  I'll give it a shot when I get home later today.  Thanks

gmelfe:
I've tried, it doesn't work. 
Similar effect as trying to resume with app_id.

lrojas94:
I get all sorts of errors when doing this too. It errors out every time and on multiple projects. Seems like it's trying to re-run all steps again (?) as from what I can tell it even rolled back one of my sampels.

Here's a sample error:
```
---------- GPT PILOT EXITING WITH ERROR ----------
Traceback (most recent call last):
  File "/mnt/d/Work/gpt-pilot/pilot/main.py", line 52, in <module>
    project.start()
  File "/mnt/d/Work/gpt-pilot/pilot/helpers/Project.py", line 134, in start
    self.developer.start_coding()
  File "/mnt/d/Work/gpt-pilot/pilot/helpers/agents/Developer.py", line 40, in start_coding
    self.implement_task(i, dev_task)
  File "/mnt/d/Work/gpt-pilot/pilot/helpers/agents/Developer.py", line 67, in implement_task
    return self.execute_task(convo_dev_task, task_steps, development_task=development_task, continue_development=True, is_root_task=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/d/Work/gpt-pilot/pilot/helpers/agents/Developer.py", line 260, in execute_task
    result = self.step_human_intervention(convo, step)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #100, opened 2023-09-21
Steps of dev are recorded in the postgres, but main.py cant continue to develop normally


lhplol:
I installed the postgres database, and the steps of my application are recorded in the database, but when I use commands to continue development, main.py will only enter the step where I describe my requirements. Please tell me how should I configure it so that the application can continue to develop normally?

![A0D9272C-FF1C-4ba8-8F1E-5B5BF207F55C](https://github.com/Pythagora-io/gpt-pilot/assets/36524051/788ac491-d441-4043-aabd-ae2067b36262)

////
(pilot-env) [lhp@localhost pilot]$ python main.py app_id=8b5ea0e5-25f7-4f16-8c1f-d7324cd66772 

------------------ LOADING PROJECT ----------------------
lhp20 (app_id=8b5ea0e5-25f7-4f16-8c1f-d7324cd66772)
--------------------------------------------------------------

? Describe your app in as many details as possible.                                                           
? How did GPT Pilot do? Were you able to create any app that works? Please write any feedback you have or just
 press ENTER to exit:  
(pilot-env) [lhp@localhost pilot]$ 
(pilot-env) [lhp@localhost pilot]$ python main.py app_id=8b5ea0e5-25f7-4f16-8c1f-d7324cd66772 step=12

------------------ LOADING PROJECT ----------------------
lhp20 (app_id=8b5ea0e5-25f7-4f16-8c1f-d7324cd66772)
--------------------------------------------------------------

Project_description already done for this app_id: 8b5ea0e5-25f7-4f16-8c1f-d7324cd66772. Moving to next step...
User_stories already done for this app_id: 8b5ea0e5-25f7-4f16-8c1f-d7324cd66772. Moving to next step...
Architecture already done for this app_id: 8b5ea0e5-25f7-4f16-8c1f-d7324cd66772. Moving to next step...
---------- GPT PILOT EXITING WITH ERROR ----------
Traceback (most recent call last):
  File "/home/lhp/gpt-pilot/pilot/main.py", line 36, in <module>
    project.start()
  File "/home/lhp/gpt-pilot/pilot/helpers/Project.py", line 83, in start
    if self.args['step'] is not None and STEPS.index(self.args['step']) < STEPS.index('coding'):
ValueError: '12' is not in list
--------------------------------------------------
? How did GPT Pilot do? Were you able to create any app that works? Please write any feedback you have or just
 press ENTER to exit:  
[...]

PR #149, opened 2023-10-08
Don't save blank app before checking progress_steps

nalbion:
Dhanushkumar pointed out that by saving the empty project too early, it was writing empty content over the top of non-empty projects.

image


Umpire2018:
image

Confirmed, Resolves #100, resolves #104, resolves #114 

[restore_console_output.txt](https://github.com/Pythagora-io/gpt-pilot/files/12842229/restore_console_output.txt)

PR #168, opened 2023-10-10
added args to get_sys_message

Dhanushkumar-S-G:
For getting System messages  `get_sys_message()` is used. In prompt `prompts/system_messages/architect.prompt` there is an variable {{ app_type }} is used but no arguments is passed.

image

So when the prompt is created, it is left blank

image

So I added arguments to `get_sys_messages()` so that it could used not only for architect, but can be used for all agents if required in future!!

### DIFF
--- a/pilot/helpers/AgentConvo.py
+++ b/pilot/helpers/AgentConvo.py
@@ -29,7 +29,7 @@ class AgentConvo:
         self.high_level_step = self.agent.project.current_step
 
         # add system message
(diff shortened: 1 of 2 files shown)


nalbion:
Would it be better to more explicitly provide `{'app_type': args['app_type']}`?

Dhanushkumar-S-G:
I thought this would more generic, so even if any variables  added to other system messages this could used without any modification

PR #170, opened 2023-10-10
Update Developer.py

vishalraj8:
This code continuously prompts the user for input, displaying a header and footer message along with feedback for invalid input. It logs the user's input and returns a dictionary with a success flag when the user enters "continue


LeonOstrez:
There are few issues with this PR.

- Changed code is part of dev_help_needed() function so indentations are wrong
- instead of input() we have to use styled_text() because it contains extra logic
- no need for extra red logs
- no real reason to refactor this part of code

Issue #121, opened 2023-10-02
Create .gpt-pilot folder


nalbion:
I like AutoGPT's `ai_settings.yaml` which is used to store, read (and manually edit) the `ai_name`, `ai_role` and `ai_goals`.

We could have a `.gpt-pilot` directory (yes, some/all of this is in the DB, but text files may make it easier to manage some of it):

- .gpt-pilot/
  - config file to save/read the initial prompt etc

See also - https://trello.com/c/ZBRiM39j

nalbion:
It would be good if the user could add/edit stories at any time and the system was able to determine if/which stories had changed (record hash of each story?)

Issue #38, opened 2023-09-04
Getting "Too many tokens in messages" value error


vijayashok99:
Full error for context when trying to create a GPT4 connection: - 
'''
  File "/Users/vijayashok/code/gpt-pilot/pilot/utils/llm_connection.py", line 94, in create_gpt_chat_completion
    raise ValueError(f'Too many tokens in messages: {tokens_in_messages}. Please try a different test.')
'''

tonydavis629:
I'm unable to use pilot for this reason. Starting from an app id causes this crash immediately, leaving no way to finish the app.

nalbion:
This is related to #84

nalbion:
I've seen in the logs where the app gets into trouble in `Developer.implement_task()`. It looks like it's currently trying to do what the Tech Lead is supposed to be doing according the the flow diagrams.

I have one request to OpenAI that includes all of the following:

- development/task/breakdown.prompt  (which is HUGE!)
- assistant replies with a detailed plan (which is also quite large)
(development/parse_task.prompt does not seem to be included in the context, it wouldn't add any value to the context.
I guess that's what `convo_dev_task.remove_last_x_messages(2)` is for)
- dev_ops/ran_command.prompt
- dev_ops/debug.prompt
- dev_ops/ran_command.prompt
- dev_ops/debug.prompt
- ...and on and on...

I think the easy fix would be updating "An item might be a code change <strike>or a command run</strike>". There's much more that needs to change in this area though.

In my scenario I asked it to build a non-trivial IntelliJ plugin. The `breakdown` prompt includes

> This should be a simple version of the app so you don't need to aim to provide a production ready code

which I feel is contradictory to the following sentence:

> Do not leave any parts of the code to be written afterwards. Make sure that all the code you provide is working and does as outlined in the description area above.

That's something that you'd _never_ tell a real developer. "Ok, hear are all of the requirements - do it all, by the end of day". And it flies in the face of what's said in the 1st paragraph of the README:

> starts coding the app step by step, like in real life while you oversee the development process. It asks you to review each task it finishes or to help when it gets stuck

The output at this stage a handful of files like:

```python
fun main(args: Array<String>) {
   println("Hello, Kotlin World!")
}

class SettingsUI() {
    // implement your Settings UI here
}

class MyToolWindow() {
   // Implement your tool window here
}

object APIHandler {
   fun handleRequest() {
      println("Handling the request")
   }
}
```

```gradle
sourceSets {
[...]

PR #175, opened 2023-10-11
create .gpt-pilot directory, save project info & chat log

nalbion:
As per #121 

This gives a much clearer view of what's going on behind the scenes. From here we can better adjust the prompts to reduce token usage (#38) and identify the source of other issues.

image

I'd like to see this `.gpt-pilot/project.yaml` become the new source for resuming projects - that way people could add/edit/remove `user_stories` etc.

image

### DIFF
--- a/pilot/helpers/Project.py
+++ b/pilot/helpers/Project.py
@@ -24,7 +24,7 @@ from utils.dot_gpt_pilot import DotGptPilot
 
 class Project:
     def __init__(self, args, name=None, description=None, user_stories=None, user_tasks=None, architecture=None,
-                 development_plan=None, current_step=None, ipc_client_instance=None):
+                 development_plan=None, current_step=None, ipc_client_instance=None, enable_dot_pilot_gpt=True):
         """
         Initialize a project.
 
@@ -70,7 +70,7 @@ class Project:
             self.architecture = architecture
         # if development_plan is not None:
         #     self.development_plan = development_plan
-        self.dot_pilot_gpt = DotGptPilot()
+        self.dot_pilot_gpt = DotGptPilot(log_chat_completions=enable_dot_pilot_gpt)
 
     def set_root_path(self, root_path: str):
         self.root_path = root_path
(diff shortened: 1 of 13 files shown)


pilot/helpers/Project.py
```diff
@@ -19,11 +19,12 @@
 from database.models.file_snapshot import FileSnapshot
 from database.models.files import File
 from logger.logger import logger
+from utils.dot_gpt_pilot import DotGptPilot
 
 
 class Project:
     def __init__(self, args, name=None, description=None, user_stories=None, user_tasks=None, architecture=None,
-                 development_plan=None, current_step=None, ipc_client_instance=None):
+                 development_plan=None, current_step=None, ipc_client_instance=None, enable_dot_pilot_gpt=True):
```
LeonOstrez, line 27: shouldn't this be False by default?

pilot/helpers/Project.py
```diff
@@ -69,6 +70,11 @@ def __init__(self, args, name=None, description=None, user_stories=None, user_ta
             self.architecture = architecture
         # if development_plan is not None:
         #     self.development_plan = development_plan
+        self.dot_pilot_gpt = DotGptPilot(log_chat_completions=enable_dot_pilot_gpt)
```
LeonOstrez, line 73: can we do this "if  enable_dot_pilot_gpt" ? Just to make sure it doesn't affect regular user at all

pilot/helpers/Project.py
```diff
@@ -69,6 +70,11 @@ def __init__(self, args, name=None, description=None, user_stories=None, user_ta
...
+
+    def set_root_path(self, root_path: str):
+        self.root_path = root_path
+        self.dot_pilot_gpt.with_root_path(root_path)
```
LeonOstrez, line 77: i would also add check if enable_dot_pilot_gpt to this line

requirements.txt
```diff
@@ -13,6 +13,7 @@ psycopg2-binary==2.9.6
 python-dotenv==1.0.0
 python-editor==1.0.4
 pytest==7.4.2
+pyyaml==6.0.1
```
LeonOstrez, line 16: why do we need this dependency now?

LeonOstrez, line 16: why do we need this dependency now?

PR #179, opened 2023-10-11
Feat: Add theme management and Unified color_text Function

Umpire2018:
This PR introduces:

1. A consolidated color_text function, improving color output management.
2. Support for the NO_COLOR environment variable, allowing users to disable colored output.


nalbion:
I think it would be better if there were semantic style names like `error`, `hint` etc

LeonOstrez:
@Umpire2018  we are using custom wrapper for print(), which is called get_custom_print(), so I believe easiest way would be to make changes in custom_print.py in local_print() add something like:

```
def get_custom_print(arguments): #changed args to arguments
...

def local_print(*args, **kwargs):
        message = " ".join(map(str, args))
        
        if 'no-color' in arguments and arguments['no-color'] is not None:
            message = remove_style(message)
```

second thing would be to remove custom_style from questionary.py if project.args['no-color'] is not None.

I think that could remove all styling but it has to be tested.

PS. you can use your approach also just have in mind you will need separate implementation for questionary in both cases.

nalbion:
I would have thought that it would make more sense to have something like `print( style_error('this is an error') )`

Whatever `style_error` does is abstracted away from the app. It may use `colorama` with red, it may use another color or no color, it may even scream it through the speakers...

Umpire2018:
> @Umpire2018 we are using custom wrapper for print(), which is called get_custom_print(), so I believe easiest way would be to make changes in custom_print.py in local_print() add something like:
> 
> ```
> def get_custom_print(arguments): #changed args to arguments
> ...
> 
> def local_print(*args, **kwargs):
>         message = " ".join(map(str, args))
>         
>         if 'no-color' in arguments and arguments['no-color'] is not None:
>             message = remove_style(message)
> ```
> 
> second thing would be to remove custom_style from questionary.py if project.args['no-color'] is not None.
> 
> I think that could remove all styling but it has to be tested.
> 
> PS. you can use your approach also just have in mind you will need separate implementation for questionary in both cases.

@LeonOstrez 

Got it, thanks.
Will it be better to move the color style from `questionary.py`  to `style.py` to make it united?

Umpire2018:
> I would have thought that it would make more sense to have something like `print( style_error('this is an error') )`
> 
> Whatever `style_error` does is abstracted away from the app. It may use `colorama` with red, it may use another color or no color, it may even scream it through the speakers...

@nalbion I see your point

`style_error` is a higher-level abstraction that utilizes color_text to display error messages, hiding the underlying details from the calling code.

Additional utility functions might include:

Warning Style
```python

def style_warning(text: str) -> str:
    return color_text(text, ColorName.YELLOW, bold=True)
```
Success Message Style
```python
def style_success(text: str) -> str:
    return color_text(text, ColorName.GREEN, bold=True)
```

But I don't think it's necessary to introduce this style_error function in this PR, as the purpose of the PR is just to implement no_color plus unified `color function`. I believe we can proceed with this in the future.

nalbion:
@Umpire2018 not too long ago we had several calls throughout the codebase directly to an external library, and the calls included extra parameters for color, bold etc, and they were specific to the external library.

These changes look like we'd be going back in that direction. I'd rather have an abstraction called throughout the code, even if (for now) it was `color_red(text)` and `color_red_bold(text)` etc which called `color_text(tex, Color.RED)`.  ...It gives us something we can search & replace

Umpire2018:
Following the aforementioned recommendations, I have submitted a new version of the code. Currently, the `style.py` file processes the `no-color` configuration within the `Config` class, and I have also introduced abstract functions based on `color_text`.


- [x] **Make sure all previous functions had been replaced by new functions.**
image

- [x] **Make sure pass tests.**
image

- [x] **Make sure `python main.py no-color` works.**

image

Umpire2018:
- [x] Light mode.

屏幕截图 2023-10-15 103907

- [x] Dark mode.
屏幕截图 2023-10-15 104120

Umpire2018:
@LeonOstrez @nalbion 
Ready for review. But i find it strange that the test did not pass. That particular section should not be the issue.

pilot/utils/style.py
```diff
@@ -1,46 +1,159 @@
...
+
+        Args:
+            text (str): The text to colorize.
+
+        Returns:
+            str: The colorized text.
+        """
+        color = style_config.get_color(color_name)
+        style = ColoramaStyle.BRIGHT if bold else ""
+        return f'{color}{style}{text}'
```
LeonOstrez, line 140: Shouldn't we reset style after each message with Style.RESET_ALL? For example before we had:

```
def red(text):
    return f'{Fore.RED}{text}{Style.RESET_ALL}'
```

nalbion:
> Ready for review. But i find it strange that the test did not pass. That particular section should not be the issue.

The tests should pass if you merge the latest `main` into your branch

pilot/utils/custom_print.py
```diff
@@ -1,4 +1,6 @@
 import builtins
+import logging
+import os
```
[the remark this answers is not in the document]
Umpire2018: Yes, you are right. Removed it.

pilot/utils/style.py
Umpire2018: #160 introduced init() for initialization. The purpose is to add the initialization of Colorama using colorama.init() at the beginning of the script.

```python
init(autoreset=True)
```
Here i use `autoreset=True`. By initializing with init(), the color auto-resets to default after every output.

Umpire2018:
> > Ready for review. But i find it strange that the test did not pass. That particular section should not be the issue.
> 
> The tests should pass if you merge the latest `main` into your branch

image
Tests failed.

Konfekt:
@LeonOstrez Thank you very much. Would it be possible to change the colors of the pair of input/output text, instead of swapping them?

immagine

Now, depending on light/dark theme, either the input or output text becomes invisible. Ideally, both would be visible.

PR #183, opened 2023-10-11
docs: Fix other arguments link

w3bdesign:
Fixes the non-working TOC link and renames it to CLI arguments as it is used further down in the README


w3bdesign:
bilde

PR #185, opened 2023-10-11
Support Anthropic, Palm, Huggingface, Ollama, TogetherAI, Replicate, etc.

krrishdholakia:
Hi @nalbion @zvone187,

Noticed you're calling the openai chat completions endpoint. I'm working on litellm (simple library to standardize LLM API Calls - https://github.com/BerriAI/litellm) and wanted to see how i could be helpful.

My PR shows users how to call their LLM providers (bedrock, togetherai, huggingface tgi, replicate, ai21, cohere, ai21 etc.) by pointing your openai base to a [local proxy](https://docs.litellm.ai/docs/proxy_server) they can use for experimentation.

This makes **no code changes** to your app.

Happy to add additional tests/documentation if the initial PR looks good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,6 +173,29 @@ python main.py app_id=<ID_OF_THE_APP> skip_until_dev_step=0
 ## `update_files_before_start`
 
 
+# Use Anthropic,Huggingface,Palm,Ollama, etc.[Full List](https://docs.litellm.ai/docs/providers)
+
+## Create OpenAI-proxy
+We'll use [LiteLLM](https://docs.litellm.ai/docs/) to create an OpenAI-compatible endpoint, that translates OpenAI calls to any of the [supported providers](https://docs.litellm.ai/docs/providers).
+
+```python
+pip install litellm
+```
+```python
+$ litellm --model ollama/codellama
+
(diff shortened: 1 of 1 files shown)


nalbion:
See also #40

krrishdholakia:
Hey @nalbion yup - looks like the last comment is also indicating usage via proxy. We've also added a tutorial for this in our docs - https://docs.litellm.ai/docs/proxy_server#tutorial-use-with-aiderautogencontinue-devlangroidetc. 

This seems like an approach that requires the least amount of code change. Thoughts?

senko:
> Hey @nalbion yup - looks like the last comment is also indicating usage via proxy. We've also added a tutorial for this in our docs - https://docs.litellm.ai/docs/proxy_server#tutorial-use-with-aiderautogencontinue-devlangroidetc.
> 
> This seems like an approach that requires the least amount of code change. Thoughts?

We now support LIteLLM proxy out of the box (via config flags): https://github.com/Pythagora-io/gpt-pilot/wiki/Using-GPT%E2%80%90Pilot-with-Local-LLMs#litellm

PR #186, opened 2023-10-12
Add `max_token` for CodeLlama

nalbion:
It _is_ possible to get `CodeLlama-34b-Instruct` to send a complete JSON, but OpenRouter does not document all of the required parameters.

OpenRouter is what the name implies - it's routing to the `togethercomputer/CodeLlama-34b-Instruct` model hosted at [TogetherAI](https://docs.together.ai/docs/inference-rest). TogetherAI's API has a `max_tokens` field. 

It might be worth switching to LiteLLM (PR #40) if it means that we can omit the middle-man and support more & customised models.

### DIFF
--- a/pilot/utils/files.py
+++ b/pilot/utils/files.py
@@ -18,12 +18,17 @@ def setup_workspace(args) -> str:
     Also creates a 'tests' folder inside the workspace.
     :param args: may contain 'root' key
     """
-    root = args.get('root') or get_parent_folder('pilot')
-    name = args.get('name', 'default_project_name')
-    project_path = create_directory(os.path.join(root, 'workspace'), name)
+    workspace = args.get('workspace')
+    if workspace:
+        project_path = workspace
+    else:
+        root = args.get('root') or get_parent_folder('pilot')
+        name = args.get('name', 'default_project_name')
+        project_path = create_directory(os.path.join(root, 'workspace'), name)
+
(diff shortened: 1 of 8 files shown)


pilot/utils/test_utils.py
```diff
@@ -3,13 +3,13 @@
 
 class TestShouldExecuteStep:
     def test_no_step_arg(self):
-        assert should_execute_step(None, 'project_description') is True
```
nalbion, line 6: @LeonOstrez I've had to modify this test. You mentioned that you've fixed an issue here - can you confirm that the updates to these tests are valid? This first one does not seem right

pilot/utils/files.py
```diff
@@ -18,12 +18,17 @@ def setup_workspace(args) -> str:
     Also creates a 'tests' folder inside the workspace.
     :param args: may contain 'root' key
     """
-    root = args.get('root') or get_parent_folder('pilot')
-    name = args.get('name', 'default_project_name')
-    project_path = create_directory(os.path.join(root, 'workspace'), name)
+    workspace = args.get('workspace')
```
nalbion, line 21: @LeonOstrez I've also restored this as it broke a unit test.

The `workspace` and `root` args were both introduced at the same time. The difference is that `workspace` can be used to select an existing project workspace, whereas `root` will create a new directory

`'{root}/workspace/{name}' == workspace`

PR #190, opened 2023-10-12
ReAct prompting in debugger

nalbion:
experimenting with ReAct prompting in debugger. may remove `criticism`


pilot/helpers/AgentConvo.py
```diff
@@ -114,6 +104,33 @@ def send_message(self, prompt_path=None, prompt_data=None, function_calls: Funct
 
         return response
 
+    def format_message_content(self, response, function_calls):
```
nalbion, line 107: @LeonOstrez @zvone187 is there any reason why we were formatting the response as text?

LeonOstrez: so we can store it in database. That is why there is "# TODO remove this once the database is set up properly"

Issue #134, opened 2023-10-05
KeyError: '\\U'


heijligers:
I just got this one.

---------- GPT PILOT EXITING WITH ERROR ----------
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/python@3.10/3.10.13/Frameworks/Python.framework/Versions/3.10/lib/python3.10/sre_parse.py", line 1051, in parse_template
    this = chr(ESCAPES[this][1])
KeyError: '\\U'

During handling of the above exception, another exception occurred:
`
Traceback (most recent call last):
  File "/Users/datascience/PycharmProjects/gpt-pilot2/pilot/main.py", line 90, in <module>
    project.start()
  File "/Users/datascience/PycharmProjects/gpt-pilot2/pilot/helpers/Project.py", line 134, in start
    self.developer.start_coding()
  File "/Users/datascience/PycharmProjects/gpt-pilot2/pilot/helpers/agents/Developer.py", line 40, in start_coding
    self.implement_task(i, dev_task)
  File "/Users/datascience/PycharmProjects/gpt-pilot2/pilot/helpers/agents/Developer.py", line 67, in implement_task
    return self.execute_task(convo_dev_task, task_steps, development_task=development_task, continue_development=True, is_root_task=True)
  File "/Users/datascience/PycharmProjects/gpt-pilot2/pilot/helpers/agents/Developer.py", line 228, in execute_task
    convo.load_branch(function_uuid)
  File "/Users/datascience/PycharmProjects/gpt-pilot2/pilot/helpers/AgentConvo.py", line 152, in load_branch
    self.replace_files()
  File "/Users/datascience/PycharmProjects/gpt-pilot2/pilot/helpers/AgentConvo.py", line 159, in replace_files
    self.replace_file_content(msg['content'], file['path'], file['content'])
  File "/Users/datascience/PycharmProjects/gpt-pilot2/pilot/helpers/AgentConvo.py", line 168, in replace_file_content
    updated_message, num_replacements = re.subn(pattern, new_section_content, message, flags=re.DOTALL)
  File "/opt/homebrew/Cellar/python@3.10/3.10.13/Frameworks/Python.framework/Versions/3.10/lib/python3.10/re.py", line 220, in subn
    return _compile(pattern, flags).subn(repl, string, count)
[...]

PR #191, opened 2023-10-12
Update AgentConvo.py

Iampd2002:
Issue #134 
**Fix -** **Escape \U in file paths to prevent KeyError in regular expressions**

This pull request addresses a bug in the replace_file_content method of the AgentConvo class. The bug was causing a KeyError when a file path containing the \U sequence was used in a regular expression. The fix involves escaping the \U sequence in the file path before it's used in the regular expression. This prevents the KeyError from being raised and allows the replace_file_content method to function as expected.

### DIFF
--- a/pilot/helpers/AgentConvo.py
+++ b/pilot/helpers/AgentConvo.py
@@ -168,6 +168,9 @@ class AgentConvo:
     def replace_file_content(self, message, file_path, new_content):
         escaped_file_path = re.escape(file_path)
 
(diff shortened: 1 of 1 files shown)


nalbion:
That's great @Iampd2002 , it should resolve #134. Could you please add a unit test to catch regressions in the future?

Don't worry about the other failing tests, they'll be fixed on another branch

heijligers:
Thanks for fixing this

PR #192, opened 2023-10-12
fix loading of project from where we left off when using app_id

LeonOstrez:


longfin (commented):
Upon reviewing this PR, I've noticed that the project structure has changed significantly since this PR was merged. The 'pilot' directory mentioned in the original changes is no longer present, and the file structure has been reorganized.

Given these changes, I cannot directly review the specific code modifications mentioned in this PR. However, based on the PR title and the files that were changed, it appears that this PR addressed an issue with loading projects from where they were left off when using an app_id.

Recommendations for future work:
1. Ensure that the functionality for loading projects with app_id is still working as expected in the new project structure.
2. Update documentation to reflect the new project structure and any changes in how app_id is used for project loading.
3. Consider adding or updating tests to verify the project loading functionality, especially with regards to app_id usage.

While I cannot provide a detailed code review due to the structural changes, the overall intent of this PR seems valuable for improving project continuity and state management.

PR #196, opened 2023-10-13
Update README.md

Deng-Xian-Sheng:
update local model readme


Umpire2018:
I believe it should not be imported at this time as the functionality has not been fully tested and the project is based on GPT-4.

PR #208, opened 2023-10-13
resume app through workspace

Dhanushkumar-S-G:
Since the userid is being created every time, resumption of the app is not possible because the user id is new

image

Also I have done finding workspace through relative path

### DIFF
--- a/pilot/database/database.py
+++ b/pilot/database/database.py
@@ -200,9 +200,9 @@ def get_app(app_id, error_if_not_found=True):
         return None
 
 
-def get_app_by_user_workspace(workspace):
+def get_app_by_user_workspace(user_id,workspace):
     try:
-        user_app = UserApps.get(UserApps.workspace == workspace)
+        user_app = UserApps.get((UserApps.user == user_id) & (UserApps.workspace == workspace))
         return user_app.app
     except DoesNotExist:
         return None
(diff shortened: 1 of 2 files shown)


nalbion:
`username_to_uuid()` _should_ return a consistent value each time it's called.

`user_id` doesn't serve much purpose if everybody is running their own database, but I'm guessing the idea of having username/password was that multiple people could share the same database, so the idea here was that each user could have a different path on their local machine to the project workspace.

Dhanushkumar-S-G:
Got it !! fixed now.

pilot/database/database.py
```diff
@@ -200,7 +200,7 @@ def get_app(app_id, error_if_not_found=True):
         return None
 
 
-def get_app_by_user_workspace(user_id, workspace):
+def get_app_by_user_workspace(user_id,workspace):
```
nalbion, line 203: restore space

```suggestion
def get_app_by_user_workspace(user_id, workspace):
```

pilot/utils/arguments.py
```diff
@@ -31,9 +31,10 @@ def get_arguments():
 
     app = None
     if 'workspace' in arguments:
-        app = get_app_by_user_workspace(arguments['user_id'], arguments['workspace'])
+        arguments['workspace'] = os.path.abspath(arguments['workspace'])
+        app = get_app_by_user_workspace(arguments['user_id'],arguments['workspace'])
```
nalbion, line 35: add space
```suggestion
        app = get_app_by_user_workspace(arguments['user_id'], arguments['workspace'])
```

nalbion (approved):

Issue #222, opened 2023-10-15
Enable debug steps to be aware of each other


nalbion:
When GPT Pilot wants to debug an issue, it sometimes runs a command which will affect the next command in the debugging process. Currently, all commands are returned at once and we want them to be returned one by one so next one can be impacted by the previous one.

nalbion:
```json
{
"steps": [
        {
            "type": "command",
            "command": {
                "command": "npm install socket.io",
                "timeout": 3000
            },
            "check_if_fixed": false
        },
        {
            "type": "command",
            "command": {
                "command": "npm start",
                "timeout": 3000,
                "command_id": "npm_start"
            },
            "check_if_fixed": true
        }
    ]
}
```

```
INFO: ---------- execute_task() step #0: {'type': 'command', 'command': {'command': 'npm install socket.io', 'timeout': 3000}, 'check_if_fixed': False}

INFO: Running `npm install socket.io` on Windows
INFO: Terminating process 38348
INFO: `npm install socket.io` took 3138ms to execute.
INFO: `npm install socket.io` exit code: None         - run_command_until_success()
INFO:   result: {'success': True, 'cli_response': 'stdout:\n```\n\n```'}       - execute_task

INFO: ---------- execute_task() step #1: {'type': 'command', 'command': {'command': 'npm start', 'timeout': 3000, 'command_id': 'npm_start'}, 'check_if_fixed': True}
```

nalbion:
Another scenario is where the LLM needs to read a file or check the output of a command to better inform subsequent steps:

```json
"steps": [
      {
        "type": "command",
        "command": {
          "command": "type package.json",
          "timeout": 3000
        },
        "check_if_fixed": false
      },
      {
        "type": "command",
        "command": {
          "command": "npm install koa-static",
          "timeout": 10000,
          "success_message": "added"
        },
        "check_if_fixed": false
      },
      {
        "type": "command",
        "command": {
          "command": "npm run start",
          "timeout": 3000,
          "success_message": "Server is running on port",
          "command_id": "complex_app_start"
        },
        "check_if_fixed": true
      },
```

Issue #198, opened 2023-10-13
gpt-pilot can't have other answer apart from 'yes'


pezzos:
gpt-pilot try to install mongodb on macos and it succeed but to try it, gpt-pilot wants to run 'mongo' instead of 'mongosh' so it failed. I didn't notice the first time so I let it do it.
Then I checked and I tried to tell to use the right  command:

```--------- EXECUTE COMMAND ----------
If yes, just press ENTER
? Can I execute the command: `mongo` with 2000ms timeout? No, please use 'mongosh' instead```

It doesn't work and gpt-pilot tried again to install it and I tried again to prevent it:

```--------- EXECUTE COMMAND ----------
If yes, just press ENTER
? Can I execute the command: `brew tap mongodb/brew` with 10000ms timeout? No, you can't because it won't work. You must use 'mongosh' instead of 'mongo'.```


Then I said no to all the next command:
```--------- EXECUTE COMMAND ----------
If yes, just press ENTER
? Can I execute the command: `brew install mongodb-community@4.4` with 30000ms timeout? no


answer: no

--------- EXECUTE COMMAND ----------
If yes, just press ENTER
? Can I execute the command: `brew services start mongodb-community@4.4` with 10000ms timeout? no


answer: no

--------- EXECUTE COMMAND ----------
If yes, just press ENTER
? Can I execute the command: `mongo` with 2000ms timeout? no```

But I don't have a way to tell gpt-pilot to use the correct bin.

When a question is raised by gtp-pilot, it should either continue with the 'yes/ENTER' or try to understand the user input to have a real feedback. 
Here, it loops and I'm stuck.

PR #223, opened 2023-10-16
Debug step awareness

nalbion:
WIP - issue #222 (https://trello.com/c/9J7l9WYQ/10-enable-steps-to-be-aware-of-each-other)

- [x] When asked for permission to run a command, user input other then "no", "yes", "skip" is sent to LLM to update the task step list  #198  

> Can I run `npm install express`?

> User: No, use koa instead

> Can I run `npm install koa`?

- [x] `ran_command` prompt includes better message when command times out.

- [x] LLM can add `need_to_see_output` and CLI response will be sent to LLM to update task step list #222 

> LLM: `{ command: 'cat package.json', need_to_see_output: true }`

> Agent: `{ dependencies: { ...`

> LLM: `{ command: 'npm install missing_dep' }`

- [x] Debugger can kill process by `command_id`
- [x] processes started by `r` no longer hang / require CTRL+C
- [x] `execute_command_and_check_cli_response()` passes on `command_id`
- [x] `components/list_running_processes.prompt` reusable process list


pilot/helpers/agents/Developer.py
```diff
@@ -137,39 +167,45 @@ def step_human_intervention(self, convo, step: dict):
                     'r': lambda conv: run_command_until_success(conv,
                                                                 self.run_command,
                                                                 command_id='app',
-                                                                timeout=None,
+                                                                timeout=1000,
```
LeonOstrez, line 170: how come you put timeout when human asks to run command?

pilot/utils/llm_connection.py
```diff
@@ -96,10 +96,14 @@ def create_gpt_chat_completion(messages: List[dict], req_type, project,
...
 
     # Advise the LLM of the JSON response schema we are expecting
+    messages_length = len(gpt_data['messages'])
     add_function_calls_to_request(gpt_data, function_calls)
 
     try:
         response = stream_gpt_completion(gpt_data, req_type, project)
+
+        # Remove JSON schema and any added retry messages
+        gpt_data['messages'] = gpt_data['messages'][:messages_length]
```
LeonOstrez, line 106: can you explain what exactly is this doing and why?

pilot/helpers/cli.py
```diff
@@ -249,12 +248,15 @@ def execute_command(project, command, timeout=None, success_message=None, comman
...
+            done_or_error_response = 'was interrupted by user'
         else:
             print('\nTimeout detected. Stopping command execution...')
             logger.warn('Timeout detected. Stopping command execution...')
+            done_or_error_response = f'took longer than {timeout}ms so I killed it'
 
-        was_success = False
         terminate_process(process.pid)
+        # update the returncode
+        process.poll()
```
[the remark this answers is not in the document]
nalbion: because otherwise `returncode` is None, which gives the false impression that it's still running

pilot/utils/llm_connection.py
nalbion: Modifying `gpt_data['messages']` didn't actually do anything, so I've changed it to pop the new messages from the list.

`add_function_calls_to_request()` and `@retry_on_exception` add extra messages to the conversation (JSON schema and validation errors) which do not add value to the context in subsequent completion requests

pilot/helpers/agents/Developer.py
nalbion: We want to wait a moment to see if it crashes or not, but if we don't specify a timeout, you have to press CTRL+C to move forward. This allows the user to move forward with the conversation while the app runs in the background.

There's still an issue with the LLM occasionally saying "check the output on the terminal when you send a request from the browser/Postman" - #236

PR #231, opened 2023-10-18
Optimize TooDeepRecursionError Handling

### LINKED ISSUE #214: After `MAX_RECUSION_LAYER`, tell the user what should they fix themselves
In `Debugger.debug()` we raise `TooDeepRecursionError()` when we get to `MAX_RECUSION_LAYER`.

This is called from `Developer.step_human_intervention()` and `test_code_changes()`.  

We should catch the exception and tell the user what they should fix.

Umpire2018:
- Improved the handling of TooDeepRecursionError to to provide clearer error messages. close #214

### DIFF
--- a/pilot/helpers/Debugger.py
+++ b/pilot/helpers/Debugger.py
@@ -22,6 +22,7 @@ class Debugger:
             command (dict, optional): The command to debug. Default is None.
             user_input (str, optional): User input for debugging. Default is None.
             issue_description (str, optional): Description of the issue to debug. Default is None.
+            is_root_task (bool, optional): Pass to `developer.execute_task`
 
         Returns:
             bool: True if debugging was successful, False otherwise.
@@ -29,8 +30,18 @@ class Debugger:
 
         self.recursion_layer += 1
         if self.recursion_layer > MAX_RECUSION_LAYER:
+            # Provide detailed error information
+            details = {
+                "Current Recursion Layer": self.recursion_layer,
+                "Max Allowed Recursion Layer": MAX_RECUSION_LAYER,
+                "Conversation": str(convo),
+                "Command": command,
+                "User Input": user_input,
+                "Issue Description": issue_description
+            }
+            error_message = "\n".join([f"{key}: {value}" for key, value in details.items()])
             self.recursion_layer = 0
-            raise TooDeepRecursionError()
+            raise TooDeepRecursionError(context=error_message)
 
         function_uuid = str(uuid.uuid4())
         convo.save_branch(function_uuid)
--- a/pilot/helpers/cli.py
+++ b/pilot/helpers/cli.py
@@ -448,7 +448,7 @@ def run_command_until_success(convo, command,
                     'command_id': command_id,
                     'success_message': success_message,
                 })
-            except TooDeepRecursionError as e:
+            except TooDeepRecursionError:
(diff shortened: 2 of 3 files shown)


pilot/helpers/cli.py
```diff
@@ -448,9 +448,9 @@ def run_command_until_success(convo, command,
                     'command_id': command_id,
                     'success_message': success_message,
                 })
-            except TooDeepRecursionError as e:
+            except TooDeepRecursionError:
                 # this is only to put appropriate message in the response after TooDeepRecursionError is raised
-                raise TooDeepRecursionError(cli_response) if return_cli_response else e
+                raise TooDeepRecursionError(cli_response)
```
LeonOstrez, line 453: we cant remove this if statement. This is how we get out of recursion

pilot/helpers/Debugger.py
```diff
@@ -22,15 +22,26 @@ def debug(self, convo, command=None, user_input=None, issue_description=None, is
...
+                "Max Allowed Recursion Layer": MAX_RECUSION_LAYER,
+                "Conversation": str(convo),
+                "Command": command,
+                "User Input": user_input,
+                "Issue Description": issue_description
+            }
+            error_message = "\n".join([f"{key}: {value}" for key, value in details.items()])
             self.recursion_layer = 0
-            raise TooDeepRecursionError()
+            raise TooDeepRecursionError(context=error_message)
```
LeonOstrez, line 44: Isn't this going to show all level errors?
For example if we go 3 levels deep, this will at the end show all 3 levels of errors, right?

What we want to do is show only initial error

pilot/helpers/cli.py
Umpire2018: The original one like:
```python 
if return_cli_response :
    raise TooDeepRecursionError(cli_response)
else:
    raise e
```

In both way, `TooDeepRecursionError` will be raised but just the text differ, if i am right.
So i move the text to
```python
class TooDeepRecursionError(Exception):
    def __init__(self, context=None, message='Recursion is too deep!'):
        if context:
            message = f"{message} (Context: {context})"
        super().__init__(message)
```
If no context is provided, will show the messgae `Recursion is too deep`.

pilot/helpers/Debugger.py
Umpire2018: I believe this will only show fourth error.
As I understand, i take that as `TooDeepRecursionError` will be raised at recursion_layer = 4, like previous handling. It will not raised when recursion_layer = 3.
I believe i haven't changed the origin logic but just add detailed info when it once shows up.
```
if self.recursion_layer > MAX_RECUSION_LAYER:
            # Provide detailed error information
            details = {
                "Current Recursion Layer": self.recursion_layer,
                "Max Allowed Recursion Layer": MAX_RECUSION_LAYER,
                "Conversation": str(convo),
                "Command": command,
                "User Input": user_input,
                "Issue Description": issue_description
            }
            error_message = "\n".join([f"{key}: {value}" for key, value in details.items()])
            self.recursion_layer = 0
            raise TooDeepRecursionError(context=error_message)
```

pilot/helpers/cli.py
LeonOstrez: I will have to test this today or if you can share logs/screenshots of testing that would be great

Issue #147, opened 2023-10-08
ProvidesContext interface


nalbion:
Project, Story, .Module/Package should implement a new ProvidesContext interface

- to_context_prompt(stage)
- to_context_documentation()
- get_context(stage)

PR #235, opened 2023-10-18
Continue dev

LeonOstrez:
add ability to continue adding features or debugging after project is finished

### DIFF
--- a/pilot/helpers/agents/Developer.py
+++ b/pilot/helpers/agents/Developer.py
@@ -35,11 +35,8 @@ class Developer(Agent):
         self.debugger = Debugger(self)
 
     def start_coding(self):
-        if self.project.finished:
-            development_plan = self.project.feature_development_plan
-        else:
+        if not self.project.finished:
             self.project.current_step = 'coding'
-            development_plan = self.project.development_plan
             update_app_status(self.project.args['app_id'], self.project.current_step)
 
             if self.project.skip_steps is None:
@@ -49,12 +46,15 @@ class Developer(Agent):
         print(color_green_bold("🚀 Now for the actual development...\n"))
         logger.info("Starting to create the actual code...")
 
-        for i, dev_task in enumerate(development_plan):
+        for i, dev_task in enumerate(self.project.development_plan):
             self.implement_task(i, dev_task)
 
         # DEVELOPMENT END
         self.project.dot_pilot_gpt.chat_log_folder(None)
         if not self.project.finished:
+            self.project.current_step = 'finished'
+            self.project.finished = True
+            update_app_status(self.project.args['app_id'], self.project.current_step)
             message = 'The app is DONE!!! Yay...you can use it now.\n'
             logger.info(message)
             print(color_green_bold(message))
@@ -80,7 +80,7 @@ class Developer(Agent):
             "array_of_objects_to_string": array_of_objects_to_string,  # TODO check why is this here
             "directory_tree": self.project.get_directory_tree(True),
             "current_task_index": i,
-            "development_tasks": self.project.feature_development_plan if self.project.finished else self.project.development_plan,
+            "development_tasks": self.project.development_plan,
             "files": self.project.get_all_coded_files(),
             "task_type": 'feature' if self.project.finished else 'app'
         })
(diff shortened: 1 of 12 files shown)


pilot/helpers/agents/Developer.py
```diff
@@ -35,26 +35,37 @@ def __init__(self, project):
...
 
-        for i, dev_task in enumerate(self.project.development_plan):
+        for i, dev_task in enumerate(development_plan):
             self.implement_task(i, dev_task)
 
         # DEVELOPMENT END
         self.project.dot_pilot_gpt.chat_log_folder(None)
-        logger.info('The app is DONE!!! Yay...you can use it now.')
-        print(color_green_bold("The app is DONE!!! Yay...you can use it now.\n"))
+        if not self.project.finished:
```
nalbion, line 57: Does `project.finished` not mean what I expect it would mean?

```suggestion
        if self.project.finished:
```

pilot/helpers/agents/Developer.py
```diff
@@ -69,8 +80,9 @@ def implement_task(self, i, development_task=None):
             "array_of_objects_to_string": array_of_objects_to_string,  # TODO check why is this here
             "directory_tree": self.project.get_directory_tree(True),
             "current_task_index": i,
-            "development_tasks": self.project.development_plan,
+            "development_tasks": self.project.feature_development_plan if self.project.finished else self.project.development_plan,
```
nalbion, line 83: Can `project.development_plan` and `project.feature_development_plan` be the same field?

pilot/helpers/agents/TechLead.py
```diff
@@ -55,4 +49,47 @@ def create_development_plan(self):
...
+                "app_summary": self.project.project_description,
+                "clarifications": self.project.clarifications,
+                "user_stories": self.project.user_stories,
+                "user_tasks": self.project.user_tasks,
+                "technologies": self.project.architecture,
+                "directory_tree": self.project.get_directory_tree(True),
+                "development_tasks": self.project.development_plan,
+                "files": self.project.get_all_coded_files(),
+                "previous_features": previous_features,
+                "feature_description": feature_description,
```
nalbion, line 71: Is this the same data that we use to create the project plan? Can we add something like `project.get_context(stage)` as I suggested in #147?

pilot/prompts/development/feature_plan.prompt
```diff
@@ -0,0 +1,79 @@
...
+{{ directory_tree }}
+```
+
+App and all its features are already finished and working. Here are files that are currently implemented:
+{% for file in files %}
+**{{ file.path }}**:
+```
+{{ file.content }}
+```
+{% endfor %}
```
nalbion, line 66: We need to add a `function`/`tool` so that the LLM can request the files that it's interested - that's out of scope for this PR, but I think there's a lot of duplication in this file which will eventually need to be addressed in both locations

nalbion (changes requested):

pilot/helpers/agents/Developer.py
LeonOstrez: project.finished is set once we are completely done with development and go to project.finish() but i see why this could be missleading, I will update it

pilot/helpers/agents/TechLead.py
LeonOstrez: There are minor differences but we should do that in future

pilot/helpers/agents/Developer.py
LeonOstrez: yes it can, i was using project.development_plan while making feature but then i removed it. Will update this

pilot/helpers/agents/Developer.py
nalbion: So `project.finished` means that the first sprint / MVP is complete, but not _finished_?

I'd expect `project.finished` to be set to true if the user presses ENTER to declare that the project is "finished".
...but even then, the user should be able to come back in a day/week/month/year and say "hey, one more thing...".

Can we change it to `if self.project.sprint == 0` or `if self.project.current_step != 'coding'`?

LeonOstrez: Yeah, you are right, project.finished means that MVP or in other words initial project is finished. Option to add more features/improvements is always there but it doesn't mean that project is not finished.

Probably down the line we will have to change that when we add new functionalities but I think we can leave it the way it is for now

PR #260, opened 2023-10-21
debugger improvements

nalbion:
with these changes I was able to move through the debugging process much more easily


pilot/helpers/Debugger.py
```diff
@@ -33,6 +34,9 @@ def debug(self, convo, command=None, user_input=None, issue_description=None, is
             self.recursion_layer = 0
             raise TooDeepRecursionError()
 
+        convo = AgentConvo(self.agent)
```
nalbion, line 37: @LeonOstrez , @zvone187 can we remove the `convo` parameter from this method?

PR #266, opened 2023-10-22
Mistral7b & gpt-3.5 support

nalbion:
Mistral7b gets through `developer.execute_task()`


joshuafsharp:
Tested apps on this branch, and they fail at `mkdir <app_name>`, which is usually the first task. Mistral almost works but returns incorrect response `DONE` , causing a loop. Looks like implementing a JSON schema to force `DONE` or `NEEDS_DEBUGGING` response from the LLM might make it more reliable. I can try to commit an update if you want?

Also noticed that the directory it's creating is nesting another of the already created app dir (`workspace/<app name>/<app name>`), so it struggles to include that second directory when trying to run the `server.js` when it's testing the file it created.

nalbion:
> Tested apps on this branch, and they fail at `mkdir <app_name>`, which is usually the first task. Mistral almost works but returns incorrect response `DONE` , causing a loop. Looks like implementing a JSON schema to force `DONE` or `NEEDS_DEBUGGING` response from the LLM might make it more reliable. I can try to commit an update if you want?
> 
> Also noticed that the directory it's creating is nesting another of the already created app dir (`workspace/<app name>/<app name>`), so it struggles to include that second directory when trying to run the `server.js` when it's testing the file it created.

It would good to add `thoughts` and `reasoning` to explain what needs debugging and why, and pass that to debugger

CRD716:
Hope this is continued in the new branch

PR #271, opened 2023-10-23
Update requirements in preparation for 3.12

CRD716:
Prepare for 3.12, fix things, etc etc, (inb4 this ruins everything)

I have specifically only updated to versions with no breaking changes, in some cases (ex. urllib3) the version I have selected is extremely out of date, but I wanted to be conservative in my updating.

Links to all not fully updated dependencies' changelogs, ordered by how out of date they still are:

1. urllib3 https://github.com/urllib3/urllib3/blob/main/CHANGES.rst (Updating to 1.26.7, latest is 2.0.7, many things have changed.)
2. yaspin https://github.com/pavdmyt/yaspin/releases (Updating to 2.5.0, latest is 3.0.1, small change list, worth investigating.)
3. peewee https://github.com/coleifer/peewee/releases (Updating to 3.16.3, latest is 3.17.0, seems like behavior has changed.)

Based on some of the changelogs I've read, this could also increase performance in some areas.


CRD716:
questionary 2.0.1 depends on prompt_toolkit<=3.0.36 and >=2.0, whoops.

CRD716:
The regression in prompt_toolkit can be traced to https://github.com/tmbo/questionary/pull/270.

> Explicitly set the prompt_toolkit version to be <=3.0.36 because 3.0.37 has introduced some event-loop-related bug.

Umpire2018:
First and foremost, thank you very much for your contribution.

Speaking of version changes, I suggest introducing modifications to pilot/utils/utils.py in this PR. The current version used by distro is 1.8.0. The usage of `distro.linux_distribution` in line `119` was deprecated in version 1.6.0. I recommend making changes. @CRD716 

Also, considering dependency management is a tricky issue, I recommend introducing a dependency management tool like Poetry to handle it. For instance, using poetry update can update the versions of the packages being used.

@LeonOstrez @nalbion

CRD716:
> Speaking of version changes, I suggest introducing modifications to pilot/utils/utils.py in this PR. The current version used by distro is 1.8.0. The usage of `distro.linux_distribution` in line `119` was deprecated in version 1.6.0. I recommend making changes. @CRD716
Should now match current recommendations.

MRigal:
You may also want to update the `README.md` to remove the hint to the psycopg2 binaries for Windows, which are now fixed and supported.

Also the tests for Linux should be able to run again

CRD716:
Same test failure as main, not pr related.

CRD716:
@nalbion ready for testing/merge :+1:

pilot/utils/utils.py
```diff
@@ -116,7 +116,7 @@ def get_os_info():
     }
 
     if os_info["OS"] == "Linux":
-        os_info["Distribution"] = ' '.join(distro.linux_distribution(full_distribution_name=True))
+        os_info["Distribution"] = ' '.join(distro.name(pretty=True))
```
senko, line 119: `distro.name` already returns a string so we shouldn't try to join it. Joining in fact splits the string to characters, then adds a space between each, like this:

```
$ python
Python 3.11.2 (main, Mar 13 2023, 12:18:29) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import distro
>>> ' '.join(distro.name(pretty=True))
'D e b i a n   G N U / L i n u x   1 2   ( b o o k w o r m )'
>>> 
```

Instead we should just use the return value, as is:

```
>>> print(distro.name(pretty=True))
Debian GNU/Linux 12 (bookworm)
```

requirements.txt
```diff
@@ -1,26 +1,26 @@
...
-prompt-toolkit==3.0.39
+peewee==3.16.3
+prompt-toolkit==3.0.36
 psycopg2-binary==2.9.9
 python-dotenv==1.0.0
 python-editor==1.0.4
 pytest==7.4.2
 pyyaml==6.0.1
-questionary==1.10.0
+questionary==2.0.1
```
senko, line 17: I am wary of upgrading major version of packages without extensive manual testing. Could we keep questionary at the latest 1.x instead (which also might mean we don't need to upgrade a few other that 2.x depends on?)

senko (changes requested):
Thanks for doing this! :pray:

Current `main` works with 3.12 (as psycopg2 was upgraded in another PR). I agree it would be good to update the other packages as well (what this PR does), except maybe bumping the major version for questionary.

I've added a few comments inline.

CRD716:
@senko Updated as per your comments :+1:

senko:
Hey @CRD716 I've just rebased and merged your changes in https://github.com/Pythagora-io/gpt-pilot/commit/36e5d6a384683d3885622759d4de920655d188d4. There were some merge conflicts which I resolved.

We've also switched to working in [development](https://github.com/Pythagora-io/gpt-pilot/tree/development) branch and periodically merging to `main` after additional testing (to minimize breaking stuff for people), so I rebased on top of that one and merged to it.

Thanks for your contribution! :pray:

PR #276, opened 2023-10-24
IGNORE_FOLDERS env variable added

bigvo:
When GPT Pilot compiles code to debug, all compilers create lots of files which shouldn't be tracked, and currently IGNORE_FOLDER is hard-coded as constant, this change will allow to include IGNORE_FOLDERS as environment variables, so constant may be kept as it is and shouldn't be extent to infinity with different folder names for different compilers. (As exception I added target folder which created by Rust compiler).

### DIFF
--- a/pilot/const/common.py
+++ b/pilot/const/common.py
@@ -34,7 +34,6 @@ IGNORE_FOLDERS = [
     'venv',
     'dist',
     'build',
-    # Used by Rust compiler
     'target'
 ] + [folder for folder in additional_ignore_folders if folder]
 
--- a/pilot/const/common.py
+++ b/pilot/const/common.py
@@ -1,3 +1,6 @@
+import os
+
+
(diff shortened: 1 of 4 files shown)


pilot/const/common.py
```diff
@@ -29,6 +34,8 @@
     'venv',
     'dist',
     'build',
-]
+    # Used by Rust compiler
```
LeonOstrez, line 37: please remove this comment

LeonOstrez (commented):
Can you also please:

- add to Readme this option
- add to .env.example commented out example. Something like this: `# IGNORE_FOLDERS=folder1,folder2`

pilot/const/common.py
bigvo: Done.

bigvo:
Done. Please have a look.

> Can you also please:
> 
> * add to Readme this option
> * add to .env.example commented out example. Something like this: `# IGNORE_FOLDERS=folder1,folder2`

Done. Please have a look.

PR #278, opened 2023-10-24
add storing of initial prompt if user allows

LeonOstrez:


CRD716:
Tests fail after this commit :(

Issue #275, opened 2023-10-23
UnicodeDecodeError


mijhaels:
![image](https://github.com/Pythagora-io/gpt-pilot/assets/67170467/14b34a75-e8fc-4eab-b895-fedd902389c8)

Scratch-project:
go to `"/home/mijhaels/Documentos/git/gpt-pilot/pilot/helpers/Project.py",` and set the line 216 to 

`file_content = open(full_path, 'r', encoding='iso-8859-1').read()`

Umpire2018:
> go to `"/home/mijhaels/Documentos/git/gpt-pilot/pilot/helpers/Project.py",` and set the line 216 to
> 
> `file_content = open(full_path, 'r', encoding='iso-8859-1').read()`

@Scratch-project I have added `utf-8` encoding in #257 , i think that's enough for most scenario, maybe need to add error handle for this.

pezzos:
I have almost the same issue:
```---------- GPT PILOT EXITING WITH ERROR ----------
Traceback (most recent call last):
  File "/Users/a.pezzotta/Dev/gpt-pilot/pilot/main.py", line 67, in <module>
    project.start()
  File "/Users/a.pezzotta/Dev/gpt-pilot/pilot/helpers/Project.py", line 134, in start
    self.developer.start_coding()
  File "/Users/a.pezzotta/Dev/gpt-pilot/pilot/helpers/agents/Developer.py", line 50, in start_coding
    self.implement_task(i, dev_task)
  File "/Users/a.pezzotta/Dev/gpt-pilot/pilot/helpers/agents/Developer.py", line 84, in implement_task
    "files": self.project.get_all_coded_files(),
  File "/Users/a.pezzotta/Dev/gpt-pilot/pilot/helpers/Project.py", line 192, in get_all_coded_files
    files = self.get_files([file.path + '/' + file.name for file in files])
  File "/Users/a.pezzotta/Dev/gpt-pilot/pilot/helpers/Project.py", line 216, in get_files
    file_content = open(full_path, 'r').read()
  File "/Users/a.pezzotta/devtools/pyenv/versions/3.10.2/lib/python3.10/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd3 in position 10: invalid continuation byte
--------------------------------------------------
```

I tried to do what @Scratch-project said but same:
[...]

PR #293, opened 2023-11-03
refactor and thoroughly test helpers/files.py

senko:
This refactors the file access and path handling to:

* make it work for binary files
* make it work on Windows

Should resolve https://github.com/Pythagora-io/gpt-pilot/issues/275

## Test

Here's how it looks in Windows.Correct (`\`) path separators are used throughout. Also, a binary file (a JPEG image that I just copied to the project for testing) can be loaded and saved.

Screenshot from 2023-11-13 17-31-09


pilot/helpers/test_Project.py
```diff
@@ -50,14 +51,10 @@ def test_get_full_path(self, file_path, file_name, expected):
         {'name': None, 'path': 'public/index.html', 'saved_to': f'{test_root}/public/index.html'},
         {'name': '', 'path': 'public/index.html', 'saved_to': f'{test_root}/public/index.html'},
         # TODO: Treatment of paths outside of the project workspace - https://github.com/Pythagora-io/gpt-pilot/issues/129
-        # {'name': '/etc/hosts', 'path': None, 'saved_to': '/etc/hosts'},
+        {'name': '/etc/hosts', 'path': None, 'saved_to': f'{test_root}/etc/hosts'},
```
senko, line 54: Basically this means that we never reach out outside the project directory (for now). Any absolute paths are treated as relative. I think this is the safest option until we come to the point where we really want to mess with the rest of the user's system (which we should do really carefully, like ask for every overwrite or maybe even read)

PR #298, opened 2023-11-04
don't terminate process that doesn't exist

LeonOstrez:


pilot/helpers/cli.py
```diff
@@ -84,32 +85,50 @@ def terminate_running_processes():
...
 
-   if platform.system() == "Windows":
-       term_proc_windows(pid)
-   else:  # Unix-like systems
-       term_proc_unix_like(pid)
 
-   for command_id in list(running_processes.keys()):
-       if running_processes[command_id][1] == pid:
-           del running_processes[command_id]
+def terminate_process(pid: int, name=None) -> None:
```
senko, line 111: If we're using `psutil`, we can use its cross-platform utilities for terminating the processes, such as `psutil.Process().kill()`.

They even have a recipe on how to terminate a process (but itself) and all of its children here: [kill_process_tree()](https://psutil.readthedocs.io/en/latest/#kill-process-tree). I have adapted it to our case as a suggestion (note I haven't actually tested it):

```python
def kill_proc_tree(pid):
    """Kill a process tree (including grandchildren)."""

    pid = int(pid)
    if pid == os.getpid():
        # Log this because this is a bug in the program
        log.error(f"Asked to kill myself, I won't")
        return

    try:
        parent = psutil.Process(pid)
        if not parent.is_running():
            raise psutil.NoSuchProcess()
    except psutil.NoSuchProcess:
        log.warning(f"No such process: {pid}")
        return

    children = parent.children(recursive=True)
    to_kill = children + [parent]

    for p in to_kill:
        try:
            p.kill()
        except psutil.NoSuchProcess:
            # we don't care if the process is no longer there
            pass
        except Exception as e:
            # we care about all the other possible errors
            log.warning(f"Error killing process {pid}: {e}", exc_info=True)

    # Wait at most 1s for everything to terminate, so we
    # can reap the children instead of making them zombies.
    psutil.wait_procs(to_kill, timeout=1)
```

pilot/helpers/agents/Developer.py
```diff
@@ -568,7 +568,7 @@ def test_code_changes(self, code_monkey, convo):
 
                 user_feedback = response['user_input']
                 if user_feedback is not None and user_feedback != 'continue':
-                    debug_success = self.debugger.debug(convo, user_input=user_feedback, issue_description=description)  # noqa
+                    self.debugger.debug(convo, user_input=user_feedback, issue_description=description)  # noqa
```
senko, line 571: Nitpicks:

* I don't think this needs a `noqa` any longer since it'll no longer produce a warning
* I would suggest we delete code instead of commenting it out (see next line); we have unlimited undo in Git if we need to find the old code

pilot/helpers/test_cli.py
```diff
@@ -8,10 +8,9 @@
 @patch("helpers.cli.subprocess")
 def test_terminate_process_not_running(mock_subprocess, mock_os):
     terminate_process(1234, 'not running')
-    if platform.system() == 'Windows':
-        mock_subprocess.run.assert_called_once_with(["taskkill", "/F", "/T", "/PID", "1234"])
-    else:
-        mock_os.killpg.assert_called_once_with(1234, 9)
+
+    mock_subprocess.run.assert_not_called()
```
senko, line 12: Note that if we'd move to use `psutil.Process.kill` (see other comment) this test would need changing to mock that instead, and to check it was called.

senko (approved):
Nice catch and fix!

I've got a couple of suggestions, feel free to use or ignore any of them :D

LeonOstrez:
Branch merged to "development" branch

PR #309, opened 2023-11-13
Run GitHub CI actions on development branch, too

senko:
This enables running GitHub actions on the `development` branch, so we can continue to have these checks on PRs that are made against that branch.


senko:
The falling test is not from this commit, and nicely showcases why it's useful to have tests running against `development` as well.

PR #311, opened 2023-11-13
show user option to type 'no' when asking to execute command

LeonOstrez:
Resolves https://github.com/orgs/Pythagora-io/projects/1/views/1?pane=issue&itemId=43965409


senko (approved):
:+1: this will be so helpful, I also didn't know that you could that until recently!

PR #314, opened 2023-11-14
Telemetry

senko:
Improve telemetry.

Also adds user settings in `~/.gpt-pilot/config.json`, which can in future be used for other things, not just telemetry config.

### DIFF
--- a/pilot/utils/exit.py
+++ b/pilot/utils/exit.py
@@ -6,8 +6,10 @@ import requests
 from helpers.cli import terminate_running_processes
 from utils.questionary import styled_text
 
+from utils.telemetry import telemetry
 
 def send_telemetry(path_id):
+
     # Prepare the telemetry data
     telemetry_data = {
         "pathId": path_id,
@@ -60,6 +62,7 @@ def ask_to_store_prompt(project, path_id):
     try:
         answer = styled_text(project, question, ignore_user_input_count=True)
         if answer == '':
+            telemetry.set("initial_prompt", init_prompt)
             response = requests.post("https://api.pythagora.io/telemetry", json=telemetry_data)
             response.raise_for_status()
     except requests.RequestException as err:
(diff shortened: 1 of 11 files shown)


pilot/utils/exit.py
```diff
@@ -84,6 +101,16 @@ def exit_gpt_pilot(project, ask_feedback=True):
 
     ask_to_store_prompt(project, path_id)
 
-    ask_user_feedback(project, path_id, ask_feedback)
+    # Ask user for contact email. If they choose not to provide it, ask for direct
+    # feedback. Both are optional.
+    if not ask_user_email(project, path_id, ask_feedback):
+        ask_user_feedback(project, path_id, ask_feedback)
+
+    # We can figure out if we're done, but can't distinguish between crash and interrupted, yet
```
LeonOstrez, line 109: we can know that in main.py and forward data (eg. exit_info), something like:

except Exception as e:
    exit_info = str(e)
    print(color_red('---------- GPT PILOT EXITING WITH ERROR ----------'))
    traceback.print_exc()
    print(color_red('--------------------------------------------------'))
    ask_feedback = False
except KeyboardInterrupt:
    exit_info = "Execution interrupted by user"
    ask_feedback = False
finally:
    if run_exit_fn:
        exit_gpt_pilot(project, ask_feedback, exit_info)
    sys.exit(0)

pilot/utils/exit.py
```diff
@@ -84,6 +101,16 @@ def exit_gpt_pilot(project, ask_feedback=True):
 
     ask_to_store_prompt(project, path_id)
 
-    ask_user_feedback(project, path_id, ask_feedback)
+    # Ask user for contact email. If they choose not to provide it, ask for direct
+    # feedback. Both are optional.
+    if not ask_user_email(project, path_id, ask_feedback):
+        ask_user_feedback(project, path_id, ask_feedback)
```
LeonOstrez, line 107: I would avoid asking him twice. I am ok with turning off feedback for now since we gathered a lot of it

LeonOstrez (approved):
One of new tests is failing so it would be good to check that

PR #318, opened 2023-11-15
Debug cli output

LeonOstrez:
With this PR every command run will be verified by LLM. Also, fixed issue where we don't get process.return_code correctly when terminating process for running command.

resolves https://github.com/orgs/Pythagora-io/projects/1/views/1?pane=issue&itemId=43967249
resolves https://github.com/orgs/Pythagora-io/projects/1/views/1?pane=issue&itemId=43969036

### DIFF
--- a/pilot/prompts/dev_ops/ran_command.prompt
+++ b/pilot/prompts/dev_ops/ran_command.prompt
@@ -1,7 +1,7 @@
-{{ additional_info }}I ran the command `{{ command }}`
-{%- if exit_code is number %}, the exit code was {{ exit_code }}{% endif %}
-{%- if error_response %}, it {{ error_response }}{% endif %}. The output was:
-
+{%- if additional_message %}{{ additional_message }}{% endif %}
+I ran the command `{{ command }}`
+{%- if exit_code is number %}, the exit code was {{ exit_code }}{% endif %}. The output was:
+{#%- if error_response %}, it {{ error_response }}{% endif %#}
 {{ cli_response }}
 
 If the command was successfully executed, respond with `DONE`. If it wasn't, respond with `NEEDS_DEBUGGING`.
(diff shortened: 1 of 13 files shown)


pilot/prompts/dev_ops/ran_command.prompt
```diff
@@ -1,7 +1,7 @@
-{{ additional_info }}I ran the command `{{ command }}`
-{%- if exit_code is number %}, the exit code was {{ exit_code }}{% endif %}
-{%- if error_response %}, it {{ error_response }}{% endif %}. The output was:
-
+{%- if additional_message %}{{ additional_message }}{% endif %}
+I ran the command `{{ command }}`
+{%- if exit_code is number %}, the exit code was {{ exit_code }}{% endif %}. The output was:
+{#%- if error_response %}, it {{ error_response }}{% endif %#}
```
senko, line 4: what's `it` here? Maybe we could do something like:

```
The output was:
{{ cli_response }}
{% if error_response %}The standard error output was:
{{ error_response }}
{% endif %}

(maybe that's what you meant)

senko (approved):
For some reason the PR didn't update when the base branch (`development`) was updated with dev_step changes :shrug: 

This looks good to me! I think one message needs to be updated (or I didn't understand it - see inline comment), but otherwise :+1: 

As a side note, Jinja is really annoying with its default whitespace rules, I have a saner setup somewhere so we can just use `{% ... %}` without adding a ton of newlines.

I should make a patch one of these days...

pilot/prompts/dev_ops/ran_command.prompt
LeonOstrez: With previous implementation error_response could be:

- 'was interrupted by user' then message in prompt would be: ', it was interrupted by user. The output was:'
- 'took longer than {timeout}ms so I killed it' then message in prompt would be: ', it took longer than {timeout}ms so I killed it. The output was:'

now I commented out error_response in prompt.

Issue #188, opened 2023-10-12
Use JSON Schema for all requests


nalbion:
After "removing" function calling, the new implementation requiring responses in JSON schema seems to be more reliable against several models including

- openai/gpt-3.5-turbo-16k
- mistralai/mistral-7b-instruct
- and apparently [most local models](https://github.com/Pythagora-io/gpt-pilot/issues/158#issuecomment-1758743897)

It seems too much to ask for an LLM to respond with `EVERYTHING_CLEAR` without getting extra spaces & `\` escaping the underscore.
They all seem quite happy responding in JSON though. 

The new implementation should also allow for an arbitrary number of `tools` (see also #140) to be made available such as:

- `add_task` - adds a task to the backlog (CRUD)
- `list_directory` - lists files in a directory
- `read_file` - reads a file (start/end line optional)
- `http_request` - sends a GET, POST etc

nalbion:
See also #266

PR #319, opened 2023-11-16
Z development

zvone187:


pilot/helpers/agents/Developer.py
```diff
@@ -239,8 +240,19 @@ def step_test(self, convo, test_command):
...
+        triple_backtick_pattern = r"```(?:\w+\n)?(.*?)```"
+        triple_match = re.search(triple_backtick_pattern, self.run_command, re.DOTALL)
+        # Pattern for single backtick
+        single_backtick_pattern = r"`(.*?)`"
+        single_match = re.search(single_backtick_pattern, self.run_command, re.DOTALL)
+
+        if triple_match:
+            self.run_command = triple_match.group(1).strip()
+        elif single_match:
+            self.run_command = single_match.group(1).strip()
```
LeonOstrez, line 255: it happened to me (very rarely) that i would get response from LLM '``\ncmd\n``'. Notice that there are only 2 ` and not 1 or 3.

This current implementation would fail for example above and give empty string as result. Why was this change done?

LeonOstrez (changes requested):
Tests are failing, check why and see inline comment

pilot/prompts/dev_ops/ran_command.prompt
```diff
@@ -1,7 +1,8 @@
 {{ additional_info }}I ran the command `{{ command }}`
-{%- if exit_code is number %}, the exit code was {{ exit_code }}{% endif %}
 {%- if error_response %}, it {{ error_response }}{% endif %}. The output was:
 
 {{ cli_response }}
 
 If the command was successfully executed, respond with `DONE`. If it wasn't, respond with `NEEDS_DEBUGGING`.
+
+Do not respond with anything other than these two keywords.
```
LeonOstrez, line 8: This will have to be fixed properly https://github.com/Pythagora-io/gpt-pilot/issues/188

pilot/helpers/agents/Developer.py
zvone187: The previous implementation would just remove 1 ` from each side. I would say that this is the reason why you would get 2 ` - the implementation in this PR is because LLMs usually return either a single line code (with 1 `) or multi line code (with 3 `)

PR #321, opened 2023-11-18
Z development

zvone187:


senko (approved):
I think these fixes are great, but as I haven't worked with these prompts much I defer to @LeonOstrez ' opinion.

I'd ask that we merge these changes using "Squash and Merge" button and an appropriate commit message (eg. "Fixing debugging prompts"), just to have things nicer in our git log)

pilot/helpers/agents/Developer.py
```diff
@@ -112,6 +112,7 @@ def implement_task(self, i, development_task=None):
                 result['completed_steps'] = task_steps[:step_index]
                 result['current_step'] = task_steps[step_index]
                 result['next_steps'] = task_steps[step_index + 1:]
+                result['current_step_index'] = step_index
 
                 convo_dev_task.remove_last_x_messages(1)
```
LeonOstrez, line 117: in Debugger.py you are removing last 2 messages before calling 'update_task.prompt' and here only last one. Not sure if that is ok, please just double check.

pilot/helpers/Debugger.py
```diff
@@ -75,11 +75,12 @@ def debug(self, convo, command=None, user_input=None, issue_description=None, is
                         result['completed_steps'] = steps[:step_index]
                         result['current_step'] = steps[step_index]
                         result['next_steps'] = steps[step_index + 1:]
+                        result['current_step_index'] = step_index
```
LeonOstrez, line 78: step_index counter starts with 0 i think and in prompt we use 
`{% for step in completed_steps %}
step #{{ loop.index }}:`

which i belive starts counting at 1.

pilot/helpers/agents/test_Developer.py
```diff
@@ -105,23 +105,34 @@ def test_implement_task_reject_with_user_input(self, mock_completion, mock_save,
...
 ```
 
-# Current Step:
-This step will not be executed. no, use a better command
+step #1:
+step_type: command_to_run: ``
+step #2:
+step_type: command_to_run: ``
+```
+Right now, we're working on the step #2:
```
LeonOstrez, line 116: same as comment above, here it says list of completed steps: #1 and #2 and then it says right now, we're working on step #2.

pilot/prompts/development/task/update_task.prompt
```diff
@@ -1,37 +1,46 @@
...
+step_type: {{ step.type }}{% if 'command' in step -%}
+command_to_run: `{{ step.command.command }}`{% endif -%}
+
+{% endfor %}
 ```
 
-Based on the Next Task Steps and my comments above for the Current Step, please create a new list of task steps to complete this task. The Current Step will need to be modified, and based on that any of the Next Task Steps may or may not still be valid.
-{% else -%}
-Based on my comments above for the Current Step, please create a new list of task steps to complete this task.
+Now, think step by step and create a new list of task steps to complete this task based on the output we got from running `{{ current_step.command.command }}`. It will either be the same as it is now or any of the Next Task Steps may or may not be needed anymore.
```
LeonOstrez, line 35: do we always have access to current_step.command.command?

LeonOstrez (approved):

pilot/helpers/agents/Developer.py
zvone187: You're right - good catch

zvone187: Btw, in general, we should almost always remove the last 2 messages when we want to remove the last call from the conversation.

pilot/helpers/Debugger.py
zvone187: oh, great catch. Will fix it

pilot/prompts/development/task/update_task.prompt
zvone187: We should have. This prompt is being used only when the command is being run while for code change or human intervention this won't be called at all - these steps will just be executed. Likely, in the future, this might change but will leave it to fail then so we can fix it appropriately and not put an if statement around it now because then, it will be harder to find the bug

PR #326, opened 2023-11-21
small refactor of updating list of steps inside task

LeonOstrez:
Fixes issue where GPT Pilot creates a debugging plan and then, after each command is run, it tries to rewrite that plan. This leads to confusion and infinite looping.

### DIFF
--- a/pilot/helpers/Debugger.py
+++ b/pilot/helpers/Debugger.py
@@ -53,6 +53,8 @@ class Debugger:
                 },
                 DEBUG_STEPS_BREAKDOWN)
 
+            completed_steps = []
+
             try:
                 while True:
                     logger.info('Thoughts: ' + llm_response['thoughts'])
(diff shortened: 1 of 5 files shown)


pilot/helpers/Debugger.py
```diff
@@ -66,19 +68,22 @@ def debug(self, convo, command=None, user_input=None, issue_description=None, is
...
+                        completed_steps = steps[:step_index+1]
+                        result['completed_steps'] = completed_steps
                         result['current_step'] = steps[step_index]
                         result['next_steps'] = steps[step_index + 1:]
-                        result['current_step_index'] = step_index
+                        result['current_step_index'] = step_index + len(completed_steps)
 
                         # Remove the previous debug plan and build a new one
                         convo.remove_last_x_messages(2)
+                        # todo before updating task first check if update is needed
```
senko, line 86: how will this be checked?

pilot/helpers/agents/Developer.py
```diff
@@ -92,38 +91,42 @@ def implement_task(self, i, development_task=None):
...
                 step_index = result['step_index']
-                result['completed_steps'] = task_steps[:step_index]
-                result['current_step'] = task_steps[step_index]
-                result['next_steps'] = task_steps[step_index + 1:]
-                result['current_step_index'] = step_index
+                completed_steps = steps[:step_index+1]
+                result['completed_steps'] = completed_steps
+                result['current_step'] = steps[step_index]
+                result['next_steps'] = steps[step_index + 1:]
+                result['current_step_index'] = step_index + len(completed_steps)
```
senko, line 117: Is this correct? Your change to `execute_task` (*Skip steps before continue_from_step*) means the step index there (`i`) will start from 0, skip over the completed steps (because of `continue_from_step`), and the counter will already include the count of all completed steps.

senko (commented):
One question and one thing that looks like a bug to me, unless I got the execution wrong.

pilot/helpers/Debugger.py
LeonOstrez: add new prompt asking LLM to reply with "NEEDS_UPDATE" if it is needed to change next steps or to reply with "DONE" if next steps are ok. In case it answers with "NEEDS_UPDATE" then we call update_task.prompt

pilot/helpers/agents/Developer.py
LeonOstrez: nice catch!

PR #327, opened 2023-11-23
skip command

LeonOstrez:
fix case when user types 'no' to run command and need_to_see_output flag is set to true.

### DIFF
--- a/pilot/helpers/agents/Developer.py
+++ b/pilot/helpers/agents/Developer.py
@@ -234,7 +234,10 @@ class Developer(Agent):
                 print(cli_response)
                 print(color_red('-------------------'))
 
-            result = {'success': llm_response == 'DONE' or llm_response == 'SKIP', 'cli_response': cli_response}
+            result = {
+                'success': llm_response in ["DONE", "SKIP"],
+                'cli_response': cli_response
+            }
             if cli_response is None:
                 result['user_input'] = llm_response
             else:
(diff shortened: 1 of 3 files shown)


pilot/helpers/agents/Developer.py
```diff
@@ -376,7 +376,7 @@ def execute_task(self, convo, task_steps, test_command=None, reset_convo=True,
 
                     logger.info('  step result: %s', result)
 
-                    if (not result['success']) or need_to_see_output:
+                    if (not result['success']) or (need_to_see_output and not ('user_input' in result and result['user_input'] == 'SKIP')):
```
senko, line 379: Nitpick: This can be replaced with `need_to_see_output and result.get("user_input") != "SKIP"` (since if `user_input` is not in `result`, `get` will return `None`, which won't be equal to `"SKIP"`).

pilot/helpers/agents/Developer.py
```diff
@@ -234,7 +234,7 @@ def step_test(self, convo, test_command):
                 print(cli_response)
                 print(color_red('-------------------'))
 
-            result = {'success': llm_response == 'DONE', 'cli_response': cli_response}
+            result = {'success': llm_response == 'DONE' or llm_response == 'SKIP', 'cli_response': cli_response}
```
senko, line 237: Nitpick: in these cases I like to use `llm_response in ["DONE", "SKIP"]`, looks a bit easier to read.

Also, whenever the items are not just `name: variable` (like  here, there's some logic), for readability I'd recommend having it in a few lines, at least to me this looks easier to parse:

```
result = {
    "success": ...,
    "cli_response": ...,
}
```

senko (approved):
Looks good, I've got a few readability nitpicks, feel free to either use or ignore.

pilot/helpers/agents/Developer.py
LeonOstrez: True, will fix

pilot/helpers/agents/Developer.py
LeonOstrez: 👍

PR #343, opened 2023-12-06
update prompts for task creation

LeonOstrez:
Resolves:

- [x] over-engineering for smaller apps
- [x] tasks complexity should be similar throughout the project
- [x] shouldn't create tasks that are testing only
- [x] fix issue when gpt pilot picks technologies that are not needed for the project


senko (approved):
Makes sense to me. I don't know if this will have any negative effects in combination with latest (code updating) prompt changes already in dev/main.

pilot/prompts/architecture/technologies.prompt
```diff
@@ -33,7 +33,7 @@ Here are user tasks that specify what users need to do to interact with "{{ name
 ```
 
 {% endif %}
-Now, based on the app's description, user stories and user tasks, think step by step and list the names of the technologies that will be used by your development team to create the app "{{ name }}". Do not write any explanations behind your choices but only a list of technologies that will be used.
```
zvone187, line 36: Not sure why is this removed but I guess this works now because of function calling.

pilot/prompts/system_messages/architect.prompt
```diff
@@ -1,10 +1,11 @@
...
-
-**Frontend**: You prefer using Bootstrap for creating HTML and CSS while you use plain (vanilla) Javascript.
-
-**Other**: From other technologies, if the project requires periodical script run, you prefer using cronjob (for making automated tasks), and if the project requires real time communication, you prefer Socket.io for web sockets
+You are an experienced software architect. Your expertise is in creating an architecture for an MVP (minimum viable products) for {{ app_type }}s that can be developed as fast as possible by using as many ready-made technologies as possible.
+
+Never list technologies that are not needed in the project. If for specific part of project you have to choose between multiple technologies, in that case, here are technologies that you prefer using:
+- Node.js,
+- MongoDB,
+- PeeWee,
```
zvone187, line 6: Have you tried creating a node.js app with Sqlite or postgres? Peewee is actually Python specific library which breaks node.js while it's trying to use it for postgres or sqlite.

pilot/prompts/architecture/technologies.prompt
LeonOstrez: This prompt was written when we didn't have function calling and it was returning string. Besides that, second big thing is that we are saying same thing multiple times, for example, bellow there is another line:

`Do not give any subjective thoughts or options, give exact technologies, for example:`

That makes it harder to understand, debug and change

pilot/prompts/system_messages/architect.prompt
LeonOstrez: Will try that later today and ping you here.

PR #355, opened 2023-12-10
Various fixes

senko:
* fix Windows file path escaping in file modifications
* be explicit about the log code page (also helps for Windows)
* improve file overwrite message

### DIFF
--- a/pilot/helpers/AgentConvo.py
+++ b/pilot/helpers/AgentConvo.py
@@ -212,14 +212,13 @@ class AgentConvo:
         return s
 
     def replace_file_content(self, message, file_path, new_content):
-        escaped_file_path = re.escape(file_path)
-
-        pattern = rf'\*\*{escaped_file_path}\*\*:\n```\n(.*?)\n```'
+        pattern = rf'\*\*{re.escape(file_path)}\*\*:\n```\n(.*?)\n```'
 
         # Escape special characters in new_content for the sake of regex replacement
         new_content_escaped = self.escape_specials(new_content)
+        file_path_escaped = self.escape_specials(file_path)
 
-        new_section_content = f'**{file_path}**\n```\n{new_content_escaped}\n```'
+        new_section_content = f'**{file_path_escaped}**\n```\n{new_content_escaped}\n```'
 
         updated_message, num_replacements = re.subn(pattern, new_section_content, message, flags=re.DOTALL)
 
--- a/pilot/helpers/Project.py
+++ b/pilot/helpers/Project.py
@@ -121,7 +121,7 @@ class Project:
                     print('yes/no', type='button')
                     should_overwrite_files = styled_text(
                         self,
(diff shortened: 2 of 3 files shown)


pilot/helpers/AgentConvo.py
```diff
@@ -212,14 +212,13 @@ def escape_specials(self, s):
         return s
 
     def replace_file_content(self, message, file_path, new_content):
-        escaped_file_path = re.escape(file_path)
-
-        pattern = rf'\*\*{escaped_file_path}\*\*:\n```\n(.*?)\n```'
+        pattern = rf'\*\*{re.escape(file_path)}\*\*:\n```\n(.*?)\n```'
```
senko, line 215: This just removes `escape_file_path` variable to avoid confusion with similarly-named but differently escaped file path a few lines below. No changes to the pattern.

pilot/helpers/AgentConvo.py
```diff
@@ -212,14 +212,13 @@ def escape_specials(self, s):
...
 
         # Escape special characters in new_content for the sake of regex replacement
         new_content_escaped = self.escape_specials(new_content)
+        file_path_escaped = self.escape_specials(file_path)
 
-        new_section_content = f'**{file_path}**\n```\n{new_content_escaped}\n```'
+        new_section_content = f'**{file_path_escaped}**\n```\n{new_content_escaped}\n```'
```
senko, line 221: Windows paths have plenty of occurrences of `\`  which were not escaped in a recent change to this.

pilot/helpers/Project.py
```diff
@@ -121,7 +121,7 @@ def start(self):
                     print('yes/no', type='button')
                     should_overwrite_files = styled_text(
                         self,
-                        f'Do you want to overwrite the dev step {self.args["skip_until_dev_step"]} code with system changes? Type y/n',
+                        f"Can I overwrite any changes that you might have made to the project since last running GPT Pilot (y/n)?",
```
senko, line 124: Unrelated to other changes, this is just a rewording of the message because the previous one is really unclear even to people who work on GPT Pilot (ie. me :smile: ).

pilot/logger/logger.py
```diff
@@ -8,7 +8,11 @@ def setup_logger():
     log_format = "%(asctime)s [%(filename)s:%(lineno)s - %(funcName)20s() ] %(levelname)s: %(message)s"
 
     # Create a log handler for file output
-    file_handler = logging.FileHandler(filename=os.path.join(os.path.dirname(__file__), 'debug.log'), mode='w')
+    file_handler = logging.FileHandler(
+        filename=os.path.join(os.path.dirname(__file__), 'debug.log'),
+        mode='w',
+        encoding='utf-8',
```
senko, line 14: Adding encoding so logging doesn't crash on Windows when there are characters which don't fit into whatever code page is active by default.

PR #356, opened 2023-12-10
Documentation

LeonOstrez:
Creation of documentation for project:

- Technical writer agent
- creation of README.md

Screenshot 2023-12-10 at 7 33 20 PM
Screenshot 2023-12-10 at 7 39 37 PM
Screenshot 2023-12-10 at 7 46 26 PM


pilot/utils/files.py
```diff
@@ -41,3 +41,17 @@ def create_directory(parent_directory, new_directory):
...
+
+def count_lines_of_code(files):
+    total_lines = 0
+
+    for file in files:
+        # Split the content into lines
+        lines = file['content'].split('\n')
+
+        # Count the lines, skipping empty lines and comments if desired
+        line_count = sum(1 for line in lines if line.strip() and not line.strip().startswith('//'))
```
senko, line 54: I don't think there's a need to skip empty lines and comments. These are usually counted in the file lengths, unless you're talking about "real code lines" which is a bit more complex to calculate than this.

This also only removes one-line JS comments, totally ignoring `#` style comments and multi-line `/* */` comments, so it's not terribly accurate.

Suggested solution:

```
return sum(len(file.splitlines()) for file in files)
```

pilot/helpers/agents/Developer.py
```diff
@@ -50,10 +50,22 @@ def start_coding(self):
...
+
+            for threshold in progress_thresholds:
+                if current_progress_percent > threshold and threshold not in documented_thresholds:
+                    self.project.technical_writer.document_project(current_progress_percent)
+                    documented_thresholds.add(threshold)
+
             self.implement_task(i, dev_task)
 
         # DEVELOPMENT END
+        self.project.technical_writer.document_project('100')
```
senko, line 68: Why is 100 here a string?

senko (approved):
Looks okay to me (I have a few minor comments, up to you if you want to change anything).

It's basically impossible to review the effects for this. In the future I think we should have  adding some logs/screenshots something to the PR description to show the effect (like you did here) a standard practice, otherwise the reviewer can just hope the prompts are okay.

Issue #337, opened 2023-12-03
TypeError: a bytes-like object is required, not 'str'


lerisse:
This error is occuring all the time during my build. I'm unable to figure out why and when this happens. I have tried advanced mode as well as regular. Currently using GTP-4.

---------- GPT PILOT EXITING WITH ERROR ----------
Traceback (most recent call last):
  File "/Users/-/Projects/JukeBox App/gpt-pilot/pilot/main.py", line 77, in <module>
    project.start()
  File "/Users/-/Projects/JukeBox App/gpt-pilot/pilot/helpers/Project.py", line 141, in start
    self.developer.start_coding()
  File "/Users/-/Projects/JukeBox App/gpt-pilot/pilot/helpers/agents/Developer.py", line 54, in start_coding
    self.implement_task(i, dev_task)
  File "/Users/-/Projects/JukeBox App/gpt-pilot/pilot/helpers/agents/Developer.py", line 101, in implement_task
    result = self.execute_task(convo_dev_task,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/-/Projects/JukeBox App/gpt-pilot/pilot/helpers/agents/Developer.py", line 380, in execute_task
    convo.load_branch(function_uuid)
  File "/Users/-/Projects/JukeBox App/gpt-pilot/pilot/helpers/AgentConvo.py", line 183, in load_branch
    self.replace_files()
  File "/Users/-/Projects/JukeBox App/gpt-pilot/pilot/helpers/AgentConvo.py", line 190, in replace_files
    self.replace_file_content(msg['content'], file['path'], file['content'])
  File "/Users/-/Projects/JukeBox App/gpt-pilot/pilot/helpers/AgentConvo.py", line 219, in replace_file_content
    new_content_escaped = self.escape_specials(new_content)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/-/Projects/JukeBox App/gpt-pilot/pilot/helpers/AgentConvo.py", line 193, in escape_specials
    s = s.replace("\\", "\\\\")
        ^^^^^^^^^^^^^^^^^^^^^^^
TypeError: a bytes-like object is required, not 'str'

senko:
Thanks for reporting this.

Did the project (where this crashed) have any files that weren't text files (such as images, zip files, etc...)?

Issue #354, opened 2023-12-10
TypeError: a bytes-like object is required, not 'str'


KevinGeLe:
GPT-Pilot (commit): 930
OS: Ubuntu 22.04
Python: 3.9

![image](https://github.com/Pythagora-io/gpt-pilot/assets/125448701/dae8db89-60f2-4913-bc5d-4488814e947d)

Wouldn't it be easier to just check if it was an object or a string. Instead of just halting?

senko:
This should always be a string, so the fact this crashed means there's a bug somewhere else in the code that calls this. Did the project (where this crashed) have non-text files (eg. images, favicons, zip files, stuff like that)?

KevinGeLe:
> This should always be a string, so the fact this crashed means there's a bug somewhere else in the code that calls this. Did the project (where this crashed) have non-text files (eg. images, favicons, zip files, stuff like that)?

It only had .js and .db from sqlite3. The database wasnt corrupt as i could navigate the tables and .js runs without error. But gpt-pilot had some problems where it did not create the .db the first few tries.

notnitsuj:
I'm having the same issue. Looking into the workspace directory, I can see that it has created some non-text files like webpack and some in node_modules. How can I deal with this?

![Screenshot from 2023-12-12 00-46-40](https://github.com/Pythagora-io/gpt-pilot/assets/75330856/6eaa0ecd-2754-404c-a012-b0a1348d32a2)

kingshah13:
me too

clickbrain:
Just got the same error after I believe 19 dev steps.

clickbrain:
Per #363 I didn't send any files to the LLM and got this message. Anything having to do with binary files was generated by. gptpilot if that was the cause, but I don't think it was.

dsitnik:
@clickbrain @senko 
I changed the function escape_specials in AgentConvo and it fixed the issue for me. I don't know what could be causing the error.
```
    def escape_specials(self, s):
        try:
            # Try decoding with utf-8
            if isinstance(s, bytes):
                s = s.decode('utf-8')
        except UnicodeDecodeError:
            try:
                # Try another encoding if UTF-8 fails
                s = s.decode('iso-8859-1')
            except UnicodeDecodeError:
                # Handle cases where decoding fails
                s = s.decode('utf-8', errors='replace')
        s = s.replace("\\", "\\\\")

        # List of sequences to preserve
        sequences_to_preserve = [
[...]

PR #363, opened 2023-12-13
Only send text files to the LLM

senko:
We only want to send text (non-binary) files to the LLM.

Aside from not making sense, it also crashes the replacement function as shown in this screenshot from one of the reported issues:

crash

Fixes: #337, #354


LeonOstrez:
i don't understand how is this going to prevent sending of non text files to LLM?

Main function for sending files to LLM is get_all_coded_files(). replace_files() that is modified in this PR is just updating content of files that are already in LLM message. Or am I missing something?

PR #406, opened 2023-12-23
Updated files.py Check path is a file before opening.

FuzTheCat:
Mandus on Discord supplied the fix with help from ChatGPT-4.

### DIFF
--- a/pilot/helpers/files.py
+++ b/pilot/helpers/files.py
@@ -1,6 +1,6 @@
 from pathlib import Path
 import os
-from typing import Optional, Union
+from typing import Optional, Union, Dict
 
 from utils.style import color_green
 
@@ -34,7 +34,7 @@ def update_file(path: str, new_content: Union[str, bytes]):
 
 def get_file_contents(
     path: str, project_root_path: str
-) -> dict[str, Union[str, bytes]]:
+) -> Dict[str, Union[str, bytes]]:
     """
     Get file content and metadata.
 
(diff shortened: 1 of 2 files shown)


LeonOstrez:
can you please make PR to "development" branch and not "main". Thank you

senko:
Rebasing on top of development branch.

pilot/helpers/files.py
```diff
@@ -1,6 +1,6 @@
 from pathlib import Path
 import os
-from typing import Optional, Union
+from typing import Optional, Union, Dict
```
senko, line 3: Using `Dict`, `List`, etc [is deprecated](https://peps.python.org/pep-0585/#implementation) and from 3.9 (which is the minimal version we support), `dict`, etc should be used

pilot/helpers/files.py
```diff
@@ -50,26 +50,35 @@ def get_file_contents(
     will be a Python string. If that fails, it will be treated as a
     binary file and `content` will be a Python bytes object.
     """
+    # Normalize the path to avoid issues with different path separators
+    full_path = os.path.normpath(os.path.join(project_root_path, path))
```
senko, line 54: Hmmm, `path` here should already be full path, adding these two together will probably not work.

pilot/helpers/files.py
```diff
@@ -50,26 +50,35 @@ def get_file_contents(
...
+
+    # Check if the path is a file
+    if not os.path.isfile(full_path):
```
senko, line 57: It'd be better to do this as a general `except Exception` below so it can catch any other errors that might happen (eg. permission error, etc).

pilot/helpers/files.py
```diff
@@ -50,26 +50,35 @@ def get_file_contents(
...
+        raise ValueError(f"The path provided is not a file: {full_path}")
+
     try:
         # Assume it's a text file using UTF-8 encoding
-        file_content = open(path, "r", encoding="utf-8").read()
+        with open(full_path, "r", encoding="utf-8") as file:
```
senko, line 62: This change has no effect.

pilot/helpers/files.py
```diff
@@ -50,26 +50,35 @@ def get_file_contents(
...
+            file_content = file.read()
     except UnicodeDecodeError:
         # If that fails, we'll treat it as a binary file
-        file_content = open(path, "rb").read()
+        with open(full_path, "rb") as file:
```
senko, line 66: This change has no effect.

pilot/helpers/files.py
```diff
@@ -50,26 +50,35 @@ def get_file_contents(
...
+            file_content = file.read()
     except FileNotFoundError:
-        raise ValueError(f"File not found: {path}")
+        raise ValueError(f"File not found: {full_path}")
 
-    file_name = os.path.basename(path)
-    relative_path = str(Path(path).parent.relative_to(project_root_path))
+    file_name = os.path.basename(full_path)
+    relative_path = os.path.relpath(os.path.dirname(full_path), project_root_path)
```
senko, line 72: This change has no effect.

senko (changes requested):
Thanks for contributing the PR! :pray:

There's a couple of things that aren't required (I guess ChatGPT has a bit different style of coding eg dict/Dict, Path/os.path). The important thing is raising ValueError so the rest of the code knows to ignore the file.

If you'd like to rework it a bit to remove extra (unneccessary) changes and just handle the opening problem (my preference would be via catching an Exception but the isfile check is also good), that would be great!

PR #463, opened 2023-12-29
Support for modifying large files

senko:
This overhauls the way we do file modifications to support larger files and more complex modifications.

Instead of asking the LLM to output entire file with any modifications applied (and dealing with the consequence of limits to how large files LLM can code in one go), here we ask the LLM to show us "before" and "after" code snippets and then do search/replace manually.


pilot/const/function_calls.py
```diff
@@ -200,29 +200,48 @@ def command_definition(description_command='A single command that needs to be ex
                             'properties': {
                                 'type': {
                                     'type': 'string',
-                                    'enum': ['command', 'code_change', 'human_intervention'],
+                                    'enum': ['command', 'save_file', 'modify_file', 'human_intervention'],
```
senko, line 203: I renamed `code_change` to `save_file` to avoid LLM being confused what `code_change` should be used for. The `modify_file` one is new.

pilot/const/function_calls.py
```diff
@@ -200,29 +200,48 @@ def command_definition(description_command='A single command that needs to be ex
...
                                             'type': 'string',
-                                            'description': 'Full content of the file that needs to be implemented. **IMPORTANT**When you want to add a comment that tells the user to add the previous implementation at that place, make sure that the comment starts with `[OLD CODE]` and add a description of what old code should be inserted here. For example, `[OLD CODE] Login route`.',
+                                            'description': 'Full content of the file that needs to be implemented. Remember, you MUST NOT omit any of the content that should go into this file.',
                                         },
                                     },
                                     'required': ['name', 'path', 'content'],
                                 },
+                                'modify_file': {
+                                    'type': 'object',
+                                    'description': 'A file that should be modified. This should only be used for large existing files (more than 100 lines of code).',
```
senko, line 228: My plea to LLM to use this only for larger files fell on deaf ears (I guess that's because LLM has no ears?). It largery gets ignored. No matter, I feel this still should make it more likely to consider so I left it in there.

pilot/const/function_calls.py
```diff
@@ -431,75 +450,6 @@ def command_definition(description_command='A single command that needs to be ex
     }
 }
 
-GET_FILES = {
```
senko, line 434: These few schema definitions were used only in the now-removed functionality atlhough naming suggests they were more generic.

pilot/database/database.py
```diff
@@ -193,7 +193,7 @@ def get_app(app_id, error_if_not_found=True):
         return app
     except DoesNotExist:
         if error_if_not_found:
-            raise ValueError(f"No app with id: {app_id}")
+            raise ValueError(f"No app with id: {app_id}; use python main.py --get-created-apps-with-steps to see created apps")
```
senko, line 196: Totally unrelated to the rest of commit but crept into - small quality of life improvement, tell people immediately how they can see the list of apps (CLI version).

pilot/helpers/AgentConvo.py
```diff
@@ -54,7 +54,7 @@ def send_message(self, prompt_path=None, prompt_data=None, function_calls: Funct
 
         # TODO: move this if block (and the other below) to Developer agent - https://github.com/Pythagora-io/gpt-pilot/issues/91#issuecomment-1751964079
         # check if we already have the LLM response saved
-        if self.agent.__class__.__name__ == 'Developer':
+        if hasattr(self.agent, 'save_dev_steps') and self.agent.save_dev_steps:
```
senko, line 57: @LeonOstrez I believe you forgot this one in your recent refactor so added it here - LMK if this is incorrect.

pilot/helpers/Project.py
```diff
@@ -226,7 +226,7 @@ def get_files(self, files):
                 _, full_path = self.get_full_file_path(file_path, file_path)
                 file_data = get_file_contents(full_path, self.root_path)
             except ValueError:
-                file_data = {"path": file_path, "content": ''}
+                file_data = {"path": file_path, "name": os.path.basename(file_path), "content": ''}
```
senko, line 229: `name` is required by `Project.save_file()` so some edge cases had this failing.

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,286 @@
-from const.function_calls import GET_FILES, IMPLEMENT_CHANGES
+import os.path
```
senko, line 1: This is a complete rewrite, pay no attention to the diff, it's best viewed as a completely new file.

Here's the core logic:

1. If we know which file we need to work on (as a result of us being called via `modify_file` in the project task breakdown, great). If not (which is the case when we're called from debugging), we need to figure out the file name from the code change description (ie. the instructions that come out of the debugging steps).
2. Next, we tell the LLM to identify all parts of the file that need changing, and for each part, provide a block of lines in the original, and the new block of lines that should replace the original.
3. There's a bunch of nuance there, for example LLM might not indent the blocks although they're indented in the original file so we need to compensante for that.
4. We try to find all possible matches for each code block. If there's exactly one match, great, we can do the replacement. If there are zero matches, we report an error. If there's more than one match, we report an error. We need to be careful with things like `});` of which there can be multiple, but indented differently, and we need to find all possible matches with all possible indentations, otherwise the replacement will go to an incorrect location.
5. If there was an error, we send a new message to the same conversation, saying there's an error and please retry. We do that up to 5 times (so 6 attempts in total to get it right). In most cases the LLM gets it after 2 or 3 retries.
6. If it's a really complex situation and LLM keeps getting it wrong, as a fallback we ask the LLM to just produce the output file (ie. fall back to the old behaviour but without the OLD CODE trick). Sometimes if there are a bunch of large changes, it's really easier to just output the entire file.

pilot/helpers/agents/Developer.py
```diff
@@ -143,46 +142,38 @@ def implement_task(self, i, development_task=None):
...
         return {"success": True}
 
+    def step_modify_file(self, convo, task_description, step, i, test_after_code_changes):
+        # TODO fix this - the problem is in GPT response that sometimes doesn't return the correct JSON structure
+        if 'modify_file' not in step:
+            data = step
+        else:
+            data = step['modify_file']
+
+        print(f'Updating existing file {data["name"]}: {data["code_change_description"].splitlines()[0]}')
```
senko, line 172: This just displays the first line of the change/instructions so it's easier to debug if you're looking at the output

pilot/helpers/agents/Developer.py
```diff
@@ -143,46 +142,38 @@ def implement_task(self, i, development_task=None):
...
-        return files_with_comments
-
     def step_code_change(self, convo, task_description, step, i, test_after_code_changes):
         if 'code_change_description' in step:
-            # TODO this should be refactored so it always uses the same function call
             print(f'Implementing code changes for `{step["code_change_description"]}`')
             code_monkey = CodeMonkey(self.project, self)
-            updated_convo = code_monkey.implement_code_changes(convo, task_description, step['code_change_description'],
-                                                               step, i)
+            updated_convo = code_monkey.implement_code_changes(convo, task_description, step['code_change_description'], step, i)
```
senko, line 149: This flow happens as a result of debugging (and has no file name/path in the step data).

pilot/helpers/agents/Developer.py
```diff
@@ -405,9 +396,12 @@ def execute_task(self, convo, task_description, task_steps, test_command=None, r
                         # if need_to_see_output and 'cli_response' in result:
                         #     result['user_input'] = result['cli_response']
 
-                    elif step['type'] == 'code_change':
+                    elif step['type'] in ['save_file', 'code_change']:
```
senko, line 399: It's proobably safe to have this as `== 'save_file'` but I'm not sure if this *could* get called as a result of debugging and still use the `code_change` name, so I allow for both.

pilot/helpers/agents/test_CodeMonkey.py
```diff
@@ -1,123 +1,345 @@
-import re
```
senko, line 1: Like with `CodeMonkey.py`, this is best viewed as a complete rewrite.

pilot/main.py
```diff
@@ -1,11 +1,15 @@
...
 import os
 
 import sys
 import traceback
-from dotenv import load_dotenv
+
+try:
+    from dotenv import load_dotenv
+except ImportError:
+    raise RuntimeError('Python environment for GPT Pilot is not completely set up: required package "python-dotenv" is missing.') from None
```
senko, line 11: Again not related but crept in somehow. This seems like the first place were we're loading Python packages that are not built in, so if the virtual environment is not correctly set up, this raises an error.

This change will still raise an error, but a bit nicer one, so that people have better idea of what's going wrong.

pilot/main.py
```diff
@@ -39,16 +43,19 @@ def init():
...
         # sys.argv.append('--ux-test=' + 'continue_development')
 
-        args = init()
-
         builtins.print, ipc_client_instance = get_custom_print(args)
 
-
         if '--api-key' in args:
             os.environ["OPENAI_API_KEY"] = args['--api-key']
+        if '--api-endpoint' in args:
```
senko, line 56: Also nto related but crept in - allow VSCode extension to specify the API endpoint.

pilot/utils/arguments.py
```diff
@@ -47,7 +47,11 @@ def get_arguments():
 
     if 'app_id' in arguments:
         if app is None:
-            app = get_app(arguments['app_id'])
+            try:
+                app = get_app(arguments['app_id'])
+            except ValueError as err:
+                print(color_red(f"Error: {err}"))
```
senko, line 53: If there's no app with this app id, show the message in red and exit nicely, without the stacktrace so it doesn't look like a bug.

senko (commented):
Faily extensive change.

Prompts are best reviewed in the list of prompts (with playground links) in the google docs.

pilot/const/function_calls.py
LeonOstrez: My findings are that any time LLM has options which are not very clear (grey zone) he will in most cases stick to one. I think these descriptions go to grey zone and here are my suggestions what to try different:

First, save_file.description from:
`'A file that needs to be created or updated (completely replaced). This should be used for new files, small files (up to 100 lines), and files with where most of the content is replaced.'`
change to:
`'A file that needs to be created or file that needs to be completely replaced. This should be used for new files, files up to 100 lines, and files with where most of the content is replaced.'`

'most of the content' is still grey zone if you ask me, so maybe adding 'more than 90% of content is replaced' or something like that will make it more clear.

Second one would be modify_file.description from:
`'A file that should be modified. This should only be used for large existing files (more than 100 lines of code).'`
change to:
`'A file that should be modified. This should only be used for existing files that have more than 100 lines of code.'`

Just a suggestion, but I would always try to give as clear instructions as possible.

pilot/const/function_calls.py
```diff
@@ -664,7 +593,7 @@ def command_definition(description_command='A single command that needs to be ex
                 },
                 'content': {
                     'type': 'string',
-                    'description': 'Full content of the documentation file that needs to be saved on the disk. **IMPORTANT**When you want to add a comment that tells the user to add the previous implementation at that place, make sure that the comment starts with `[OLD CODE]` and add a description of what old code should be inserted here. For example, `[OLD CODE] Login route`.',
+                    'description': 'Full content of the documentation file that needs to be saved on the disk.',
```
LeonOstrez, line 596: not sure if removing OLD_CODE requires change in how documentation is created.

pilot/helpers/agents/Developer.py
```diff
@@ -143,46 +142,38 @@ def implement_task(self, i, development_task=None):
...
-
-        data = self.replace_old_code_comments([data])[0]
+            data = step['save_file']
 
         self.project.save_file(data)
         # TODO end
         return {"success": True}
 
+    def step_modify_file(self, convo, task_description, step, i, test_after_code_changes):
+        # TODO fix this - the problem is in GPT response that sometimes doesn't return the correct JSON structure
```
LeonOstrez, line 166: if we are using function calling it should always be correct JSON. Inside stream_gpt_completion we check if JSON is ok and if not we repeat request

pilot/helpers/agents/TechnicalWriter.py
```diff
@@ -42,11 +42,7 @@ def create_readme(self):
             "files": self.project.get_all_coded_files(),
         }, GET_DOCUMENTATION_FILE)
 
-        changes = self.project.developer.replace_old_code_comments([llm_response])
-
-        for file_data in changes:
-            self.project.save_file(file_data)
-
+        self.project.save_file(llm_response)
```
LeonOstrez, line 45: i believe that with these new changes this has to go through codeMonkey and not be saved directly but double check.

pilot/main.py
```diff
@@ -87,4 +94,3 @@ def init():
     finally:
         if run_exit_fn:
             exit_gpt_pilot(project, ask_feedback)
-        sys.exit(0)
```
LeonOstrez, line 90: why is this removed?

pilot/prompts/development/identify_files_to_change.prompt
```diff
@@ -0,0 +1,22 @@
...
+
+```
+{{ file['content'] }}
+```
+{% endfor %}
+
+Instructions:
+{{ code_changes_description }}
+
+Output *ONLY* the file path, relative to project root, in a single Markdown code block,
```
LeonOstrez, line 18: it would be much better to do this in function calling rather then directly in prompt. Two reasons:

1. when using function calling we check if output is correct json and if not we repeat request which means we should always get correct response format
2. we already have issues with similar prompts where different LLMs answer in different ways, so for consistency it would be better if we always use function calling

pilot/prompts/development/update_file.prompt
LeonOstrez, line 1: not sure if this is empty file commited for reason or mistake

pilot/test_main_e2e.py
```diff
@@ -1,3 +1,4 @@
+import os
```
LeonOstrez, line 1: why is this import if no other changes in file?

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,353 @@
...
+            return file_content
+
+        return blocks[0]
+
+
+    def identify_files_to_change(self, code_changes_description: str, files: list[dict]) -> list[str]:
+        """
+        Identify files to change based on the code changes description
+
+        We really just want one file, but this can handle mutliple files
```
LeonOstrez, line 222: typo 'mutliple'

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,353 @@
...
+                # This should never happen, but let's not crash if it does
+                logger.warning(f"No relevant files found for code change: {code_changes_description}")
+                return convo
+
+            # In practice, we always have exactly one file to change, so we can just pick the first in
+            # the list.
+            step['path'] = os.path.dirname(files_to_change[0])
+            step['name'] = os.path.basename(files_to_change[0])
+
+        rel_path, abs_path = self.project.get_full_file_path(step['path'], step['name'])
```
LeonOstrez, line 57: indentation of this line and next for loop seems wrong since it has same indentation as `else` bellow

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,353 @@
...
-        convo.remove_last_x_messages(2)
+            "full_output": False,
+            "standalone": standalone,
+            "code_changes_description": code_changes_description,
+            "file_content": file_content,
+            "file_name": file_name,
+            "files": files,
+        })
+
+        exchanged_messages = 2
```
LeonOstrez, line 81: why are we using `exchanged_messages` and not `convo.messages.length`? Is there difference?

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,353 @@
...
+        replace_complete_file = False
+
+        # Allow for up to 2 retries
+        while exchanged_messages < 7:
+            # Modify a copy of the content in case we need to retry
+            content = file_content
+
+            if re.findall('(old|existing).+code', llm_response, re.IGNORECASE):
```
LeonOstrez, line 89: will this also trigger if there is function like this inside of file?
```
fnName(old_code):
    // some logic
    new_code = old_code + 'random'
    return new_code
```

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,353 @@
...
+
+        if replace_complete_file:
+            content = self.replace_complete_file(
+                convo,
+                standalone,
+                code_changes_description,
+                file_content,
+                file_name, files
+            )
+            blocks = self.get_code_blocks(llm_response)
```
LeonOstrez, line 166: why do we call `self.get_code_blocks(llm_response)` ? I don't see where is this `blocks` used

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,353 @@
...
+        :param code_changes_description: description of the code changes
+        :param files: list of files to send to the LLM
+        :return: list of files to change
+        """
+        convo = AgentConvo(self)
+        llm_response = convo.send_message('development/identify_files_to_change.prompt', {
+            "code_changes_description": code_changes_description,
+            "files": files,
+        })
+        blocks = self.get_code_blocks(llm_response)
```
LeonOstrez, line 235: is it a problem here if `blocks` is odd length, meaning response was not as expected and is not processed properly?

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,353 @@
...
+                    "error": (
+                        "You must not omit any code from NEW_CODE. "
+                        "Please don't use coments like `// .. existing code goes here`."
+                    )
+                })
+                exchanged_messages += 2
+                continue
+
+            # Split the response into pairs of old and new code blocks
+            blocks = self.get_code_blocks(llm_response)
```
LeonOstrez, line 100: what if 2 files need to be changed but instead of 4 blocks we parse 2 blocks(wrong LLM response)? it would pass next 2 checks

pilot/main.py
senko: Unrelated change that crept in.

This call to `sys.exit()` has no effect, since there's nothing after "finally", and any exceptions that might have been raised are catched and handled before.

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,353 @@
...
+        # to modify so we need to figure that out first.
+        if 'path' not in step or 'name' not in step:
+            files_to_change = self.identify_files_to_change(code_changes_description, files)
+
+            if len(files_to_change) == 0:
+                # This should never happen, but let's not crash if it does
+                logger.warning(f"No relevant files found for code change: {code_changes_description}")
+                return convo
+
+            # In practice, we always have exactly one file to change, so we can just pick the first in
```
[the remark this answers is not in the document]
senko: Old prompts were quite ambiguous whether it can be one or more files so when I started writing that code I wasn't sure what needed to be done. In all my testing I've never had this handling more than one file. I also tweaked the prompt to ask for one file, not multiple.

I'll be changing the prompt to use JSON call so we'll force the LLM to always spew out exactly one.

pilot/helpers/agents/CodeMonkey.py
senko: indentation is correct. `for` has an `else` block.

pilot/helpers/agents/CodeMonkey.py
senko: I can't find `convo.messages.length`, maybe you meant `len(convo.messages)` ?

Here I'm counting how many messages I need to undo from the conversation. Conversation has more messages than that (since it's getting passed from Developer), so this cleans it up.

I don't know if that's actually important for the rest of the code, but I've seen other places do the same (just with a fixed number of messages), so I'm doing it here as well.

pilot/helpers/agents/CodeMonkey.py
senko: Aside from that fact that LLM would not us give complete file (for these two files that need to be changed)  because we asked it for old + new code blocks, it would also fail the third check because the new version of the first file would not be found in the old file.

pilot/helpers/agents/CodeMonkey.py
LeonOstrez: yeah, my bad. I mean `len(convo.message)`

Not sure if this is what you need, but for handling conversation messages when there is more complex logic we have:
```
convo.save_branch(uuid)
//some logic
convo.load_branch(uuid)
```
but if this works fine, it's ok.

pilot/const/function_calls.py
senko: I tweaked it so it works with the new flow (but there's no need for postprocessing because README is small enough not to trigger OLD CODE).

pilot/helpers/agents/TechnicalWriter.py
senko: This saves file correctly (as in - what we get from GPT gets saved correctly).

Though in theory that could have a problem of LLM not sending a complete file, since there's no other constraints and we always ask for full rewrite, I don't think we need to involve CodeMonkey here.

pilot/prompts/development/identify_files_to_change.prompt
senko: Switched to JSON in 8e301fa1cee720bd789e977e630c8adf55676bb1

pilot/prompts/development/update_file.prompt
senko: This was from before I figured out where I'll put the new functionality, removed.

pilot/test_main_e2e.py
senko: Removed in 8e301fa1cee720bd789e977e630c8adf55676bb1

pilot/helpers/agents/CodeMonkey.py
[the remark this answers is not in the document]
senko: We now force exactly one file through JSON response in 8e301fa1cee720bd789e977e630c8adf55676bb1

pilot/helpers/agents/CodeMonkey.py
senko: Reworded entire docstring in 8e301fa1cee720bd789e977e630c8adf55676bb1

pilot/helpers/agents/CodeMonkey.py
senko: This was accidentially left after I refactored that code to call a function (which handles it already). Removed in 8e301fa1cee720bd789e977e630c8adf55676bb1

pilot/helpers/agents/CodeMonkey.py
senko: GitHub doesn't let me know what this was referring to, but I believe this was for determining which file we need to change, which is now using JSON in 8e301fa1cee720bd789e977e630c8adf55676bb1

senko:
Last commit adds tracing of file update errors in CodeMonkey:

## Old block not found in original file

```json
{
    "event": "trace-codemonkey-file-update-error",
    "data": {
        "error": "replace-errors",
        "llm_response": "1 ## Change\nOld:\n```\ntwo\n```\nNew:\n```\nbar\n```\n",
        "details": [
            [
                1,
                "Old code block not found in the original file:\n```\ntwo\n```\nOld block *MUST* contain the exact same text (including indentation, empty lines, etc.) as the original file in order to match."
            ]
        ]
    }
}
```

## Old block has multiple matches

```json
{
    "event": "trace-codemonkey-file-update-error",
    "data": {
        "error": "replace-errors",
        "llm_response": "## Change\nOld:\n```\ntwo\n```\nNew:\n```\nbar\n```\n",
        "details": [
            [
                1,
                "Old code block not found in the original file:\n```\ntwo\n```\nOld block *MUST* contain the exact same text (including indentation, empty lines, etc.) as the original file in order to match."
            ]
        ]
    }
}
```

## Fallback to complete file update

```json
{
    "event": "trace-codemonkey-file-update-error",
    "data": {
        "error": "fallback-complete-replace",
        "llm_response": "1 ## Change\nOld:\n```\ntwo\n```\nNew:\n```\nbar\n```\n"
    }
}
```

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,406 @@
...
+                file_content = f['content']
+                break
+        else:
+            # If we didn't have the match (because of incorrect or double use of path separators or similar), fallback to directly loading the file
+            try:
+                file_content = get_file_contents(abs_path, self.project.root_path)['content']
+                if isinstance(file_content, bytes):
+                    # We should never want to change a binary file, but if we do end up here, let's not crash
+                    file_content = "... <binary file, content omitted> ..."
+            except ValueError:
```
LeonOstrez, line 50: Should we catch anything else?

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,406 @@
...
+                # File doesn't exist, we probably need to create a new one
+                file_content = ""
+
+        file_name = os.path.join(rel_path, step['name'])
+        return file_name, file_content
+
+    def implement_code_changes(
+        self,
+        convo: Optional[AgentConvo],
+        task_description: str,
```
LeonOstrez, line 60: Looks like this is not used

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,406 @@
...
+        code_changes_description: str,
+        step: dict[str, str],
+        step_index: int = 0,
```
LeonOstrez, line 63: Looks like this is not used

pilot/prompts/development/parse_task.prompt
```diff
@@ -1,10 +1,43 @@
...
+
+**IMPORTANT**: You must provide full information (including code samples, if any) from the previous message in `code_change_description` of `modify_file`, so that the developer can correctly implement the change.  For `save_file`, you MUST include FULL file contents, without omitting anything or adding comments like `// rest of the code goes here`.
+
+Examples:
+------------------------example_1---------------------------
+```
+{
+  "tasks": [
+    {
+      "type": "modify_file",
```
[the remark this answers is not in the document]
senko: fixed in 881e863e100aed971f1e055e13d2dfea6c40e649

pilot/prompts/development/parse_task.prompt
```diff
@@ -1,10 +1,43 @@
...
-When you think about in which file should the new code go to, always try to make files as small as possible and put code in more smaller files rather than in one big file. Whenever a file becomes too large (more than 50 lines of code) split it into smaller files.
+Ok, now, take your previous message that starts with `{{ instructions_prefix }}` and ends with `{{ instructions_postfix }}` and convert it to a list of actionable steps that will be executed by a machine. Analyze the entire message, think step by step and make sure that you don't omit any information when converting this message to steps.
+
+Each step can be either:
+
+* `command` - command to run (must be able to run on a {{ os }} machine, assume current working directory is project root folder)
+* `save_file` -  create new or update ONE existing file; use this if the existing file is smaller than 20 lines or if many lines need to be changed
+* `modify_file` - update ONE existing file; use this if the existing file is larger than 20 lines and only a few lines need to be updated
+
+If the step if of type `save_file` or `modify_file`, it needs to contain instructions on how to change only ONE file.
```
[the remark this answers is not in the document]
senko: fixed in 881e863e100aed971f1e055e13d2dfea6c40e649

pilot/const/function_calls.py
```diff
@@ -431,74 +450,25 @@ def command_definition(description_command='A single command that needs to be ex
...
-                'files': {
-                    'type': 'array',
-                    'description': 'List of files that need to be analized to implement the reqired changes. Any file name in this array MUST be from the directory tree listed in the previous message.',
-                    'items': {
-                        'type': 'string',
-                        'description': 'A single file name that needs to be analized to implement the reqired changes. Remember, this is a file name with path relative to the project root. For example, if a file path is `{{project_root}}/models/model.py`, this value needs to be `models/model.py`. This file name MUST be listed in the directory from the previous message.',
-                    }
+                'file': {
+                    'type': 'string',
+                    'description': 'Path to the file that needs to be modified, relative to the project root.',
```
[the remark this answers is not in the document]
senko: I'd avoid going into that particular rabbit hole right now (we should revisit all the prompts and make the LLM always use relative paths and get rid of the hacky function to guess what it meant... one day!)

pilot/prompts/development/implement_changes.prompt
```diff
@@ -1,26 +1,116 @@
...
+```
+{{ file_content }}
+```
+This file needs to be modified by these instructions:
+---------------------instructions------------------------------
+{{ code_changes_description }}
+----------------------end_of_instructions-----------------------------
+{% if full_output %}
+I want you to implement the instructions and show me the COMPLETE NEW VERSION of this file.
+**IMPORTANT**: Your reply should not omit any code in the new implementation or substitute anything with comments like `// .. rest of the code goes here ..`, because I will overwrite
```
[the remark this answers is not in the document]
senko: fixed in 49b3c4664dc732d0f540d73511b6b937ceca7340

(btw `\n`s don't cost more, we don't actually *need* to put everything in one line...)

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,406 @@
-from const.function_calls import GET_FILES, IMPLEMENT_CHANGES
+import os.path
+import re
+from typing import Optional
+
 from helpers.AgentConvo import AgentConvo
 from helpers.Agent import Agent
+from helpers.files import get_file_contents
+from logger.logger import logger
```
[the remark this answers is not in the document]
senko: fixed in bebd63c8221ef8481ad144565e7201c3ad2f2fc3

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -1,38 +1,406 @@
...
+from const.function_calls import GET_FILE_TO_MODIFY
 
+from utils.exit import trace_code_event
```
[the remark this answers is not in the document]
senko: yeah that file has a lot of things in it that aren't exit-related....

PR #465, opened 2023-12-29
don't open files when loading project

LeonOstrez:
resolves https://github.com/orgs/Pythagora-io/projects/2/views/1?pane=issue&itemId=48526799
- Remove "Saving file" and "Update file" logs
- Don't trigger "openFile" in extension when loading app

### DIFF
--- a/pilot/helpers/Project.py
+++ b/pilot/helpers/Project.py
@@ -260,7 +260,7 @@ class Project:
         path = data['path'] if 'path' in data else name
 
         path, full_path = self.get_full_file_path(path, name)
-        update_file(full_path, data['content'])
+        update_file(full_path, data['content'], project=self)
         if full_path not in self.files:
             self.files.append(full_path)
 
@@ -410,7 +410,7 @@ class Project:
 
         clear_directory(self.root_path, IGNORE_FOLDERS + self.files)
         for file_snapshot in file_snapshots:
-            update_file(file_snapshot.file.full_path, file_snapshot.content)
+            update_file(file_snapshot.file.full_path, file_snapshot.content, project=self)
             if file_snapshot.file.full_path not in self.files:
                 self.files.append(file_snapshot.file.full_path)
 
--- a/pilot/helpers/Project.py
+++ b/pilot/helpers/Project.py
@@ -386,7 +386,8 @@ class Project:
         development_step, created = DevelopmentSteps.get_or_create(id=development_step_id)
 
         for file in files:
-            print(color_cyan(f'Saving file {file["full_path"]}'))
+            if not self.check_ipc():
+                print(color_cyan(f'Saving file {file["full_path"]}'))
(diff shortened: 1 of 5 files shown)


pilot/helpers/Project.py
```diff
@@ -447,15 +448,18 @@ def ask_for_human_intervention(self, message, description=None, cbs={}, convo=No
...
                 'type': MESSAGE_TYPE[message_type],
                 'content': str(text),
             })
             if message_type == MESSAGE_TYPE['user_input_request']:
                 return self.ipc_client_instance.listen()
+        else:
+            print(text)
+
+    def check_ipc(self):
+        return self.ipc_client_instance is not None and self.ipc_client_instance.client is not None
```
LeonOstrez, line 462: Have in mind that before on all places in code we were checking
`self.ipc_client_instance is None or self.ipc_client_instance.client is None`
and now we are checking the opposite.

pilot/helpers/agents/Developer.py
```diff
@@ -465,11 +465,11 @@ def continue_development(self, iteration_convo, last_branch_name, continue_descr
             user_description = 'Can you check if the app works please? ' + user_description
 
             if self.run_command:
-                if self.project.ipc_client_instance is None or self.project.ipc_client_instance.client is None:
+                if self.project.check_ipc():
```
LeonOstrez, line 468: logic is same as before just instead of checking
`if not self.project.check_ipc():`
I swapped it around so that we are checking 
`if self.project.check_ipc():`

pilot/utils/questionary.py
```diff
@@ -28,14 +28,14 @@ def styled_text(project, question, ignore_user_input_count=False, style=None, hi
             print(color_yellow_bold(f'{user_input.user_input}'))
             return user_input.user_input
 
-    if project.ipc_client_instance is None or project.ipc_client_instance.client is None:
+    if project.check_ipc():
```
LeonOstrez, line 31: logic is same as before just instead of checking
`if not self.project.check_ipc():`
I swapped it around so that we are checking 
`if self.project.check_ipc():`

pilot/helpers/agents/Developer.py
```diff
@@ -223,10 +223,10 @@ def step_human_intervention(self, convo, step: dict):
                 self.get_run_command(convo)
 
             if self.run_command:
-                if (self.project.ipc_client_instance is None or self.project.ipc_client_instance.client is None):
-                    human_intervention_description += color_yellow_bold('\n\nIf you want to run the app, just type "r" and press ENTER and that will run `' + self.run_command + '`')
-                else:
+                if self.project.check_ipc():
```
LeonOstrez, line 226: logic is same as before just instead of checking
`if not self.project.check_ipc():`
I swapped it around so that we are checking 
`if self.project.check_ipc():`

senko (approved):
:+1: LGTM, ship it!

PR #492, opened 2024-01-04
fixes

LeonOstrez:
fix when ipc connection breaks and when get_created_apps_with_steps() throws

### DIFF
--- a/pilot/helpers/ipc.py
+++ b/pilot/helpers/ipc.py
@@ -38,5 +38,6 @@ class IPCClient:
     def send(self, data):
         serialized_data = json.dumps(data, default=json_serial)
         data_length = len(serialized_data)
-        self.client.sendall(data_length.to_bytes(4, byteorder='big'))
-        self.client.sendall(serialized_data.encode('utf-8'))
+        if self.client is not None:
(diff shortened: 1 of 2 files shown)


pilot/helpers/ipc.py
```diff
@@ -38,5 +38,6 @@ def listen(self):
     def send(self, data):
         serialized_data = json.dumps(data, default=json_serial)
         data_length = len(serialized_data)
-        self.client.sendall(data_length.to_bytes(4, byteorder='big'))
-        self.client.sendall(serialized_data.encode('utf-8'))
+        if self.client is not None:
```
LeonOstrez, line 41: when ipc connection breaks this prevents crash

pilot/main.py
```diff
@@ -60,7 +62,6 @@ def init():
                                 f"{'' if len(app['development_steps']) == 0 else app['development_steps'][-1]['id']:3}"
                                 f"  {app['name']}" for app in get_created_apps_with_steps()))
 
-            run_exit_fn = False
```
LeonOstrez, line 63: if get_created_apps_with_steps() throws exception then we will not set run_exit_fn = False so we will run exit fn when we shouldnt

senko (approved):

Issue #505, opened 2024-01-07
[Bug]: Dosn't listen to user inputs


Bengan1:
### Version

VisualStudio Code extension

### Operating System

Windows 10

### What happened?

I have a problem where the pilot wants to use the mongo path, but it dosn's work so I have told it.
`
In MongoDB version 4.4 and later, the mongo shell has been deprecated and is replaced by the new MongoDB Shell, mongosh. So use mongosh instead of mongo.
`

But it dosnt listen, its just looping around in circles trying to find the path to mongo with diffrent means, so basicly I'm stuck.

Bengan1:
Also the same then it wants to do a curl input, always the wrong format. Then I write, can you write this instead: Invoke-RestMethod -Uri "http://localhost:3000/healt"
But no response, just saying that the original command dosn't work, And then the infinite loop again.

PR #513, opened 2024-01-10
Allow user to enter input/help while debugging a problem

senko:
Allows user to specify what's wrong when starting the debugging loop:

"Can I debug" prompt:

Screenshot from 2024-01-10 10-46-30

User's message sent to gpt-pilot core:

Screenshot from 2024-01-10 10-46-46

Contents of the message sent (from `debug.prompt`, note that the previous message has the full stderr log):

```
Ok, we need to debug this issue and we need to be able to execute `npm install express dotenv` successfully. .
Here is a brief explanation of what's happening:
\```
the package.json file is corrupted and needs to be fixed
\```
The project directory tree looks like:

/: package.json


I want you to create a list of steps that are needed to debug this issue.

A step can be either a `command` or `code_change`.

`command` step will run a command on the machine and will return the CLI output to you so you can see what to do next. Note that the command will be run on a Linux machine.

`code_change` step will change the code and you need to thoroughly describe what needs to be implemented. I will implement the requested changes and let you know.

Also, make sure that at least the last step has `check_if_fixed` set to TRUE.

**IMPORTANT**
When you think about in which file should the new code go to, always try to make files as small as possible and put code in more smaller files rather than in one big file. Whenever a file becomes too large (more than 50 lines of code) split it into smaller files.
```

Fixes: #505

### DIFF
--- a/pilot/helpers/Debugger.py
+++ b/pilot/helpers/Debugger.py
@@ -47,7 +47,7 @@ class Debugger:
 
             if ask_before_debug or i > 0:
                 print('yes/no', type='button')
-                answer = ask_user(self.agent.project, 'Can I start debugging this issue?', require_some_input=False)
+                answer = ask_user(self.agent.project, 'Can I start debugging this issue [Y/n/error details]?', require_some_input=False)
                 if answer.lower() in NEGATIVE_ANSWERS:
                     return True
                 if answer and answer.lower() not in AFFIRMATIVE_ANSWERS:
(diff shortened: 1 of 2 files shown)


pilot/helpers/Debugger.py
```diff
@@ -47,7 +47,7 @@ def debug(self, convo, command=None, user_input=None, issue_description=None, is
 
             if ask_before_debug or i > 0:
                 print('yes/no', type='button')
-                answer = ask_user(self.agent.project, 'Can I start debugging this issue?', require_some_input=False)
+                answer = ask_user(self.agent.project, 'Can I start debugging this issue [Y/n/error details]?', require_some_input=False)
```
LeonOstrez, line 50: I would put same capitalisation:
`Y/N/Error details`
or
`y/n/error details`

LeonOstrez (commented):
Besides minor comment ready for merge

pilot/helpers/Debugger.py
senko: Reason for this capitalisation is that Y is default

PR #521, opened 2024-01-11
Fix/argparse for cli

### LINKED ISSUE #138: refactor `arguments.py` to use `argparse`
`argparse` makes it easier to provide help/documentation on the command line and also provides `choices`, `default` and `type`

It also returns an object rather than a dict. There are currently arg options scattered throughout the code which are not mentioned in `arguements.py`

Yugma2605:
Fixes #138 

I have added the argparse for parsing through the passed arguments. 

I have also changed Project.py, custom_print.py accordingly to prevent any errors that may cause. The code is flexible enough if some more arguments needs to be passed.

### DIFF
--- a/pilot/utils/arguments.py
+++ b/pilot/utils/arguments.py
@@ -34,6 +34,7 @@ def get_arguments():
         app = get_app_by_user_workspace(arguments.user_id, arguments.workspace)
         if app is not None:
             arguments.app_id = app.id
+            arguments.continuing_project = True
     else:
         arguments.workspace = None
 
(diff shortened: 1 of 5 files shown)


pilot/utils/arguments.py
```diff
@@ -9,82 +10,80 @@
...
-    if 'password' not in arguments:
-        arguments['password'] = 'password'
-
-    if 'step' not in arguments:
-        arguments['step'] = None
+    # Return the parsed arguments
 
     return arguments
 
+def get_email():
```
shism2, line 77: It looks like there are two get_emails()

Borda (approved):
yes, this looks much easier to read :)

LeonOstrez:
This is handled properly with version 0.2

PR #524, opened 2024-01-13
keyboard interrupt

LeonOstrez:


senko (approved):
:+1: 

In the future, it would be great if you could add description for PRs :-)

PR #563, opened 2024-01-21
Determine relevant files for each task and only send these

senko:
**Work in progress, do not merge**


senko:
Superseedd by #827

PR #565, opened 2024-01-21
backwards compatibility

LeonOstrez:
Support of backwards compatibility.

- loading of project is not done by replaying whole project anymore, but skipping to last possible step
- removed usage of `user_inputs` and `command_runs` tables in db

### DIFF
--- a/pilot/helpers/agents/Developer.py
+++ b/pilot/helpers/agents/Developer.py
@@ -1,6 +1,7 @@
 import platform
 import uuid
 import re
+import json
 
 from const.messages import WHEN_USER_DONE
 from utils.style import (
@@ -25,7 +26,7 @@ from helpers.AgentConvo import AgentConvo
 from utils.utils import should_execute_step, array_of_objects_to_string, generate_app_data
 from helpers.cli import run_command_until_success, execute_command_and_check_cli_response, running_processes
 from const.function_calls import FILTER_OS_TECHNOLOGIES, EXECUTE_COMMANDS, GET_TEST_TYPE, IMPLEMENT_TASK, COMMAND_TO_RUN
-from database.database import save_progress, get_progress_steps, update_app_status
+from database.database import save_progress, get_progress_steps, update_app_status, get_all_app_development_steps
 from utils.utils import get_os_info
 
 ENVIRONMENT_SETUP_STEP = 'environment_setup'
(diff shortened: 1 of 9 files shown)


pilot/helpers/agents/Developer.py
```diff
@@ -87,36 +106,80 @@ def implement_task(self, i, development_task=None):
...
+            remove_last_x_messages = 2
+
         steps = response['tasks']
-        convo_dev_task.remove_last_x_messages(2)
+        convo_dev_task.remove_last_x_messages(remove_last_x_messages)
 
         completed_steps = []
 
         while True:
+            # This whole if statement is loading of project.
```
senko, line 153: Should this be before (outside) of the while loop? As I understand it, it only needs to execute once (if loading)

pilot/helpers/agents/Developer.py
```diff
@@ -87,36 +106,80 @@ def implement_task(self, i, development_task=None):
...
+                # run_command is command to run app (used in continue_development)
+                if self.run_command is None:
+                    self.run_command = next(
+                        (el for el in reversed(self.project.dev_steps_to_load) if
+                         'get_run_command.prompt' in el.get('prompt_path', '')), None)
+                    if self.run_command is not None:
+                        self.run_command = json.loads(self.run_command['llm_response']['text'])['command']
+
+                self.project.cleanup_list('dev_steps_to_load', last_iteration['id'])
+                self.project.last_iteration = self.project.dev_steps_to_load[0]
```
senko, line 181: Should we have `self.project.finish_loading()` here? Or are we implicitly trusting that we're in a loop and the next iteration of tle loop won't find the last iteration because we just handled it, and it will call finish loading for us?

pilot/helpers/agents/Developer.py
```diff
@@ -281,14 +345,18 @@ def task_postprocessing(self, convo, development_task, continue_development, tas
...
         #       What is it at step_human_intervention()?
-        self.get_run_command(convo)
-
-        if development_task is not None:
-            convo.remove_last_x_messages(2)
-            detailed_user_review_goal = convo.send_message('development/define_user_review_goal.prompt', {
-                'os': platform.system()
-            }, should_log_message=False)
-            convo.remove_last_x_messages(2)
+        if self.project.last_iteration is None:
```
senko, line 348: shoulnd't these be:

```
if self.run_command is None:
   ...
if self.last_detailed_user_review_goal is None:
   ..
```

pilot/helpers/agents/Developer.py
```diff
@@ -365,6 +433,10 @@ def execute_task(self, convo, task_description, task_steps, test_command=None, r
         convo.save_branch(function_uuid)
 
         for (i, step) in enumerate(task_steps):
+            # This means we are still loading the project
+            if self.project.last_iteration is not None:
```
senko, line 437: Should we use the same check (like `if self.skip_dev_steps`)? everywhere so we don't end up in some weird case where one part of the code thinks we've finished loading and the other still thinks we're loading?

pilot/utils/questionary.py
```diff
@@ -18,6 +18,7 @@ def styled_select(*args, **kwargs):
 
 
 def styled_text(project, question, ignore_user_input_count=False, style=None, hint=None):
+    project.finish_loading()
```
senko, line 21: :dizzy_face:

senko (commented):
Bunch of questions

senko (commented):
Left a few comments. Also, when loading an app, *after* load, for a normal dev step (modify file), there's this error:

```
  File "/tmp/gpt-pilot/pilot/helpers/AgentConvo.py", line 72, in send_message
    save_development_step(self.agent.project, prompt_path, prompt_data, self.messages, response)
  File "/tmp/gpt-pilot/pilot/database/database.py", line 293, in save_development_step
    'previous_step': project.checkpoints['last_development_step']['id'],
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
TypeError: 'DevelopmentSteps' object is not subscriptable
```

PR #572, opened 2024-01-23
Loop report

LeonOstrez:
track tasks progress and report if it goes in loop

Screenshot 2024-01-23 at 2 11 59 PM

### DIFF
--- a/pilot/utils/task.py
+++ b/pilot/utils/task.py
@@ -1,0 +1,147 @@
+import json
+from uuid import uuid4
+
+from utils.telemetry import telemetry
+from utils.exit import trace_code_event
+from const.telemetry import LOOP_THRESHOLD
+
+
+class Task:
+    """
+    Task data structure to store information about the current task. The task data structure is sent to telemetry.
+    Currently used to trace big loops in the code.
+
+    >>> from utils.task import Task
+
+    To set up a new task:
+
+    >>> task = Task()
+
+    To set a value:
+
+    >>> task.set('task_description', 'test')
+
+    To increment a value:
+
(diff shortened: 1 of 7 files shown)


pilot/utils/task.py
```diff
@@ -0,0 +1,147 @@
...
+from utils.exit import trace_code_event
+from const.telemetry import LOOP_THRESHOLD
+
+
+class Task:
+    """
+    Task data structure to store information about the current task. The task data structure is sent to telemetry.
+    Currently used to trace big loops in the code.
+
+    >>> from utils.task import Task
```
senko, line 14: :heart:

pilot/utils/task.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def inc(self, key: str, value: int = 1):
+        """
+        Increment a value in the task data
+
+        :param key: key to increment
+        :param value: value to increment by
+        """
+        self.data[key] += value
+        if key == 'steps' and self.data[key] == LOOP_THRESHOLD + 1:
```
senko, line 75: Looks like this would send telemetry for *every* task in that loop over the threshold, instead of just the first one?

pilot/utils/task.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __init__(self):
+        self.initial_data = {
+            'task_description': '',
+            'task_number': 0,
+            'steps': 0,
+            'debugging': [],
+        }
+        self.data = self.initial_data.copy()
+        self.ping_extension = True
```
senko, line 56: This should probably check if we have IPC (to avoid printing in the console).

PR #573, opened 2024-01-23
Prompt updates

zvone187:


pilot/prompts/components/project_tasks.prompt
```diff
@@ -61,3 +61,7 @@ Here is an example of good and bad task description:
     "description": "In 'route.js' add a route for /dashboard that returns the status 200. Route should be accessible only for logged in users. In 'middlewares.js' there should be a check if user is logged in using session. If user is not logged in, it should redirect to /login. If user is logged in, it should return the user data. User data should be fetched from database in 'users' collection using the user id from session.",
     "user_review_goal": "Reviewer should be able to navigate to '/dashboard' page without error. If user is logged in it should show user data, if not it should redirect to /login."
 }
+
+
+Rule #5
+When you're thinking about what tasks should be first, always priritize frontend tasks to be implemented before backend tasks so that, once we get to the backend implementation, user can test the backend by clicking around the frontend and not use tools like cUrl or Postman to make API requests to test the backend.
```
LeonOstrez, line 67: typo `priritize` -> `prioritize`
suggestion `..., user can ...`  -> `..., the user can ...`

pilot/prompts/dev_ops/debug.prompt
```diff
@@ -8,7 +8,6 @@ Ok, we need to debug this issue
 ```
 {{ user_input }}
 ```{% endif -%}
-{{ context }}
```
LeonOstrez, line 11: Why is this removed?

If we don't need it anymore we should remove it from Debugger.py where we call the prompt:
```
            llm_response = convo.send_message('dev_ops/debug.prompt',
                {
                    'command': command['command'] if command is not None else None,
                    'user_input': user_input,
                    'issue_description': issue_description,
                    'os': platform.system(),
                    'context': convo.to_context_prompt()
                },
                DEBUG_STEPS_BREAKDOWN)
```

pilot/prompts/development/implement_changes.prompt
```diff
@@ -112,4 +115,40 @@ Rule #4:
 You must output the CURRENT_CODE exactly as it is in the original file, including the indentation from the original code, as it will be used for search-replace, and it should only match the original file in ONE place.
 In the NEW_CODE, remember to follow the same coding style that is used in the rest of the file. Pay special attention to the indentation of the new code and make sure to include all the required old and new code, without omitting anything.
 Pay very close attention to parenthesis and make sure that when CURRENT_CODE is replaced with NEW_CODE there are no extra parenthesis or any parenthesis missing.
+
+Rule #5:
+Pay attention to the lines of code that you select in the CURRENT_CODE and the lines that you add to NEW_CODE because all lines under CURRENT_CODE will be replaced with the lines under NEW_CODE. Sometimes you might duplicated the currently implemented lines by not putting them under CURRENT_CODE but repeating them under NEW_CODE - you must **NOT** do that.
```
LeonOstrez, line 120: typo `duplicated` -> `duplicate`

pilot/prompts/development/implement_changes.prompt
```diff
@@ -112,4 +115,40 @@ Rule #4:
...
+Here is an example. Let's say that you have a code like this:
+```
+let someClass = new SomeClass();
+let methodValue = someClass.someMethod();
+```
+And you want to add a line `methodValue += 22;` below the line `let methodValue = someClass.someMethod();`. Here is an example of what you must **NOT** do:
+------------------------incorrect example---------------------------
```
LeonOstrez, line 127: suggestion, I would add `START of incorrect example` but it can stay this way

pilot/prompts/development/implement_changes.prompt
```diff
@@ -112,4 +115,40 @@ Rule #4:
...
+```
+let someClass = new SomeClass();
+let methodValue = someClass.someMethod();
+methodValue += 22;
+```
+END
+------------------------END incorrect example---------------------------
+
+See how the line `let someClass = new SomeClass();` was added under NEW_CODE but it wasn't mentioned under CURRENT_CODE so the result will be that this line will be duplicated in the end result once the line `let methodValue = someClass.someMethod()` gets replaced with lines under NEW_CODE. Instead, this would be a correct way to do it:
+------------------------correct example---------------------------
```
LeonOstrez, line 142: same suggestion as above

pilot/prompts/development/iteration.prompt
```diff
@@ -45,4 +45,4 @@ When you think about in which file should the new code go to, always try to make
 You do not need to make any automated tests work.
 
 **IMPORTANT**
-Do not tell me anything about setting up the database or anything OS related - only if some dependencies need to be installed.
+You must not tell me to run a command in the database or anything OS related - only if some dependencies need to be installed. If there is a need to run an OS related command, specifically tell me that this should be labeled as "Human Intervention" and explain what does the human need to do.
```
LeonOstrez, line 48: suggestion `what does the human need to do.` -> `what the human needs to do.`

pilot/prompts/development/parse_task.prompt
```diff
@@ -5,6 +5,7 @@ Each step can be either:
 * `command` - command to run (must be able to run on a {{ os }} machine, assume current working directory is project root folder)
 * `save_file` -  create new or update ONE existing file; use this if the existing file is smaller than 20 lines or if many lines need to be changed
 * `modify_file` - update ONE existing file; use this if the existing file is larger than 20 lines and only a few lines need to be updated
+* `human_intervention` - if you need the human to do something, use this type of step and explain in details what you want the human to do.
```
LeonOstrez, line 8: you called it `human_intervention` but this prompt is using function call `IMPLEMENT_TASK` which is expecting result `human_intervention_description`. It should be changed here in prompt or in function call + code. I think first option is much easier.

pilot/prompts/development/task/breakdown.prompt
```diff
@@ -35,4 +35,7 @@ You do not need to make any automated tests work.
 {% endif %}
 DO NOT specify commands to create any folders or files, they will be created automatically - just specify the relative path to each file that needs to be written.
 
+**IMPORTANT**
+You must not tell me to run a command in the database or anything OS related - only if some dependencies need to be installed. If there is a need to run an OS related command, specifically tell me that this should be labeled as "Human Intervention" and explain what does the human need to do.
```
LeonOstrez, line 39: same comment as above for same prompt. Also, it might be good to have it in `components` folder so in future there is only 1 place where we have to make changes.

pilot/prompts/development/parse_task.prompt
zvone187: good catch - I changed the function call because it's more consistent with other types

pilot/prompts/development/implement_changes.prompt
zvone187: duplicated is correct

pilot/prompts/dev_ops/debug.prompt
zvone187: This context doesn't give any relevant info to the LLM

PR #581, opened 2024-01-27
Increase max timeout for command run

senko:
This is useful for for slower commands like downloads, react init, etc.


LeonOstrez (commented):
I think if we set max timeout so high we have to either give user option to stop command mid execution or give user a way to manually change timeout.

This could get user stuck for 3min waiting for very simple commands like `npm run start`

PR #582, opened 2024-01-27
Handle openai errors

senko:
Instead of crashing with an ugly stack trace, we now nicely show the error to the user. We still exit tho, this doesn't add any additional recovery mechanisms. We also still log this as a failure (of type `failure:api-error`) and include crash diagnostics so we can see any anomalies if the happen.

1. Timeout error:
Screenshot from 2024-01-27 13-26-23

2. Request too large (context size / token limit error):
Screenshot from 2024-01-27 13-20-33

3. Error when the LLM response cannot be correctly parsed:
Screenshot from 2024-01-27 13-42-18

4. Generic error from the API:
Screenshot from 2024-01-27 13-00-31

Note: The API key error should never happen because we first check the key/endpoint before starting the app (the check was disabled in this instance for testing purposes), but this will catch any generic error from the API (including SSL/connection errors).

### DIFF
--- a/pilot/main.py
+++ b/pilot/main.py
@@ -23,7 +23,7 @@ from database.database import database_exists, create_database, tables_exist, cr
 
 from utils.settings import settings, loader
 from utils.telemetry import telemetry
-
+from helpers.exceptions import ApiError, TokenLimitError
 
 def init():
     # Check if the "euclid" database exists, if not, create it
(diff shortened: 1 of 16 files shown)


pilot/main.py
```diff
@@ -103,11 +103,15 @@ def init():
...
                 print('Exit', type='exit')
 
+    except (ApiError, TokenLimitError) as err:
+        telemetry.record_crash(err, end_result="failure:api-error")
+        telemetry.send()
+        run_exit_fn = False
+        print('Exit', type='exit')
     except KeyboardInterrupt:
         telemetry.set("end_result", "interrupt")
-        if project.check_ipc():
```
LeonOstrez, line 108: why did you remove this if statement?

senko: added it back

PR #583, opened 2024-01-27
human intervention fix

LeonOstrez:
refactor function calling for IMPLEMENT_TASK and DEBUG_STEPS_BREAKDOWN.

All functionalities should stay the same, it just ensures that step type "human_intervention" has required field "human_intervention_description"

### DIFF
--- a/pilot/const/function_calls.py
+++ b/pilot/const/function_calls.py
@@ -40,6 +40,141 @@ def return_array_from_prompt(name_plural, name_singular, return_var_name):
     }
 
 
+def dev_step_type_description():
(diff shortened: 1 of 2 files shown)


pilot/const/function_calls.py
```diff
@@ -250,61 +342,14 @@ def command_definition(description_command='A single command that needs to be ex
                 "properties": {
                     "tasks": {
                         'type': 'array',
-                        'description': 'List of smaller development steps that need to be done to complete the entire task.',
+                        'description': 'List of smaller development steps.',
```
senko, line 345: do we need this change?

senko (approved):

PR #607, opened 2024-02-01
Human intervention

LeonOstrez:
Fix tracking recursion layer in Debugger and update prompt instructions on how to use Human Intervention.

### DIFF
--- a/pilot/const/function_calls.py
+++ b/pilot/const/function_calls.py
@@ -147,11 +147,11 @@ def step_human_intervention_definition():
         "properties": {
             "type": {
                 "const": "human_intervention",
-                "description": dev_step_type_description()
+                "description": 'Development step that will be executed by a human. You should avoid using this step if possible, task does NOT need to have "human_intervention" step.'
             },
             "human_intervention_description": {
                 "type": "string",
-                "description": "Description of step where human intervention is needed."
+                "description": "Very clear description of step where human intervention is needed."
             }
         },
         "required": ["type", "human_intervention_description"]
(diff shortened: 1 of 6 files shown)


pilot/const/function_calls.py
```diff
@@ -147,11 +147,11 @@ def step_human_intervention_definition():
         "properties": {
             "type": {
                 "const": "human_intervention",
-                "description": dev_step_type_description()
+                "description": 'Development step that will be executed by a human. You should avoid using this step if possible, task does NOT need to have "human_intervention" step.'
```
senko, line 150: This sounds a bit wierd, but if you tested it and it works, :shipit:

pilot/helpers/agents/Developer.py
```diff
@@ -529,6 +532,7 @@ def execute_task(self, convo, task_steps, test_command=None, reset_convo=True,
                         response = self.should_retry_step_implementation(step, step_implementation_try)
                         if 'retry' in response:
                             # TODO we can rewind this convo even more
+                            step_implementation_try += 1
```
senko, line 535: Is this *ever* increased except (now) in TokenLimitError?

Looks like it's always passed as 0 if we din't hit this specific error? Quick grep didn't find it anywhere else:

```
$ rg step_implementation_try
helpers/agents/Developer.py
405:    def should_retry_step_implementation(self, step, step_implementation_try):
406:        if step_implementation_try >= MAX_COMMAND_DEBUG_TRIES:
485:            step_implementation_try = 0
529:                        response = self.should_retry_step_implementation(step, step_implementation_try)
```

PR #609, opened 2024-02-01
Refactor modify-file to always output entire file and review it after

senko:
This does two things:

1. removes error-prone "OLD CODE BLOCK" / "NEW CODE BLOCK" approach which turns out to be too complex for even GPT4, and always creates a full new file instead
2. adds a review step that checks every proposed change and approves/rejects them (so that changes that incorrectly delete or destroy working code are identified and dropped)

Here's example output from the reviewer ("hunks" means one change in a unified diff):
```json
{
    "hunks": [
        {
            "number": 1,
            "decision": "apply",
            "reason": "This change correctly implements the logic to handle and display messages and responses from loaded conversations, aligning with the task instructions."
        },
        {
            "number": 2,
            "decision": "ignore",
            "reason": "This hunk is simply formatting changes (removal of whitespace and adjustment of indentation) without functional impact. Applying or ignoring it is a matter of code style preference."
        }
    ],
    "review_notes": "The provided changes accurately reflect the instructions for enhancing the 'Load conversation' functionality. However, it appears there's a single hunk of changes, so the review format seems to expect more. To clarify, only the modifications directly related to loading and displaying the messages and responses for conversations were necessary and correctly adjusted. No additional hunks were required or detailed in the provided diff. The review schema suggests multiple hunks might be present or expected, which might lead to confusion based on this specific scenario."
}
```

Tests TDB (currently failing)


pilot/helpers/agents/CodeMonkey.py
```diff
@@ -296,118 +170,180 @@ def identify_file_to_change(self, code_changes_description: str, files: list[dic
...
+
+        llm_response = convo.send_message('development/review_changes.prompt', {
+            "instructions": instructions,
+            "file_name": file_name,
+            "old_content": old_content,
+            "hunks": hunks,
+        }, REVIEW_CHANGES)
+
+        hunk_ids = set()
+        for hunk in llm_response.get("hunks", []):
```
LeonOstrez, line 207: before doing this for loop i would add a check if `len(hunks) == len(llm_response.get("hunks", []))` and if not i would retry send_message()

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -87,32 +90,20 @@ def implement_code_changes(
...
+            file_content,
+            file_name, files
+        )
 
+        # Review the changes and only apply changes that are useful/approved
+        if content and content != file_content:
+            content = self.review_change(convo, code_changes_description, file_name, file_content, content)
+
+        # If we have changes, update the file
         if content and content != file_content:
```
LeonOstrez, line 107: here i think we have to check if content == file_content and if yes (we didn't apply any changes) we should retry step with new changes. This could be a bit too complex for this PR so we can make new task for that

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -296,118 +170,180 @@ def identify_file_to_change(self, code_changes_description: str, files: list[dic
...
+        fallback: str
+    ):
+        """
+        Apply the diff to the original file content.
+
+        This uses the internal `_apply_patch` method to apply the
+        approved diff hunks to the original file content.
+
+        If patch apply fails, the fallback is the full new file content
+        with all the changes applied (as if the reviewer approved eveythng).
```
LeonOstrez, line 271: typo `eveythng` -> `everything`

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -296,118 +170,180 @@ def identify_file_to_change(self, code_changes_description: str, files: list[dic
...
+
+        FIXME: For verification, it also calls command-line "Patch"
+        utility to check that the internal algorithm works correctly.
+
+        :param file_name: name of the file being modified
+        :param old_content: old file content
+        :param hunks: change hunks from the unified diff
+        :param fallback: proposed new file content (with all the changes applied)
```
LeonOstrez, line 279: maybe it's safer that we fallback to original_file and once we implement retrying of step in case there are no file changes approved (i mentioned in comment above case when content == file_content) then this fallback should be same

LeonOstrez (commented):
implement_code_changes() function can have only 2 arguments, currently third one is step and second one is step.code_changes_description. Also in description of that same function we should remove
```
        :param task_description: description of the task
        :param step_index: index of the step to implement
```

pilot/helpers/agents/CodeMonkey.py
LeonOstrez: also add trace

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -296,118 +170,180 @@ def identify_file_to_change(self, code_changes_description: str, files: list[dic
...
+        """
+        Apply the diff to the original file content.
+
+        This uses the internal `_apply_patch` method to apply the
+        approved diff hunks to the original file content.
+
+        If patch apply fails, the fallback is the full new file content
+        with all the changes applied (as if the reviewer approved eveythng).
+
+        FIXME: For verification, it also calls command-line "Patch"
```
LeonOstrez, line 273: remove

PR #637, opened 2024-02-07
EXPERIMENTAL - add syntax check to reviewer

senko:
**This is experimental, do not merge.**

Adds a syntax check to reviewer.

Uses `prettier` for JS, `ruff` for Python and `gofmt` for Go. You'll need to have the tool(s) installed in the proper environment (doesn't do any checks).

Also doesn't send the files back to the codemonkey for fixup if the syntax is incorrect. As I said, it's experimental.


thewshii:
Never heard of go, this'll be fun. I should be set up bout to run it.

PR #641, opened 2024-02-11
Experimental: Add project template for Node/Express/Mongo/Boostrap

senko:
Screenshot from 2024-02-10 16-59-28

There is only one template for now: node+express+mongoose+ejs+bootstrap.


LeonOstrez:
trigger trace event when template is used

PR #646, opened 2024-02-13
Added model override for agents.

gravypower:
In an effort to help pilot become more cost effective I wanted to try and have some tasks completed by other models, GPT 3.5 Turbo for coding as an example. 

This PR introduced new env vars 

```
# Override model for an agent 
#FULL_STACK_DEVELOPER_MODEL_NAME=gpt-4-turbo-preview
#CODE_MONKEY_MODEL_NAME=gpt-4-turbo-preview
#ARCHITECT_MODEL_NAME=gpt-4-turbo-preview
#PRODUCT_OWNER_MODEL_NAME=gpt-4-turbo-preview
#TECH_LEAD_MODEL_NAME=gpt-4-turbo-preview
#DEV_OPS_MODEL_NAME=gpt-4-turbo-preview
#TECHNICAL_WRITER_MODEL_NAME=gpt-4-turbo-preview
DEFAULT_MODEL_NAME=gpt-4-turbo-preview
```

I have not tested yet with the OpenAI API, wanted to get feed back before doing this.

On a side note I found some of the agent object relationships a bit all over the place and wanted to know what your plan is for this area. I think it would benefit from making interactions more standard taking advantage of polymorphism. as an example moving logic in the AgentConvo class into the Agent base class. Would love to know your thoughts.

Thanks

### DIFF
--- a/pilot/.env.example
+++ b/pilot/.env.example
@@ -10,8 +10,18 @@ AZURE_ENDPOINT=
 OPENROUTER_API_KEY=
 
 # In case of Azure/OpenRouter endpoint, change this to your deployed model name
-MODEL_NAME=gpt-4-turbo-preview
 MAX_TOKENS=8192
+
+# Override model for an agent 
+#FULL_STACK_DEVELOPER_MODEL_NAME=gpt-4-turbo-preview
+#CODE_MONKEY_MODEL_NAME=gpt-4-turbo-preview
+#ARCHITECT_MODEL_NAME=gpt-4-turbo-preview
+#PRODUCT_OWNER_MODEL_NAME=gpt-4-turbo-preview
+#TECH_LEAD_MODEL_NAME=gpt-4-turbo-preview
+#DEV_OPS_MODEL_NAME=gpt-4-turbo-preview
+#TECHNICAL_WRITER_MODEL_NAME=gpt-4-turbo-preview
+DEFAULT_MODEL_NAME=gpt-4-turbo-preview
+
 
 # Folders which shouldn't be tracked in workspace (useful to ignore folders created by compiler)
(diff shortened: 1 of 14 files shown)


gravypower:
will address linting issues soon.

senko:
Hey @gravypower this is cool!

Yeah the current code is all over the place, it basically organically grows from what we try to make it more capable and less error prone, and as the patterns change, the code gets left behind. Our thinking for the moment is if we go in and fix/refactor, the same will happen in a couple of weeks as we discover new/different ways of setting up the conversations and agents.

senko:
WRT the multi-model approach: I do think it's going to be a challenge making it work reliably. It also might need different prompts for different models (this is a known issue for local models too, and something we've been running away from for some time :sweat_smile: )

gravypower:
I will be happy to help out when you decide to refactor. in the mean time I will fix the failing build that i missed till now.

gravypower:
> WRT the multi-model approach: I do think it's going to be a challenge making it work reliably. It also might need different prompts for different models (this is a known issue for local models too, and something we've been running away from for some time 😅 )

yeah I have been seeing a bunch of people on discord trying different models, maybe I will look at giving the ability to override prompts also? I think making the agents polymorphic can help attack the multi-model approach as we will be able to use liskov substitution to our advantage.

senko:
An idea I had previously but never tried is to allow custom prompts (root) directory, and always check there first and then fallback to the original location if not (kindof how Django resolves templates, if you've ever used it).

(this might be what you're already thinking about)

gravypower:
I am going to start with a fresh branch as I am behind. will merge in my changes there and finish the other ones.

PR #670, opened 2024-02-20
propagate list of features through all prompts

LeonOstrez:


pilot/prompts/development/feature_plan.prompt
```diff
@@ -1,19 +1,7 @@
 You are working in a software development agency and a project manager and software architect approach you telling you that you're assigned to add new feature to an existing project. You are working on a {{ app_type }} called "{{ name }}" and you need to create a detailed development plan so that developers can start developing the new feature.
 
 {{ project_details }}
-
-Here are development tasks that specify what was already coded:
-```
-{{ development_tasks }}
```
senko, line 7: is removing this intentional?

pilot/prompts/development/plan.prompt
```diff
@@ -1,7 +1,7 @@
 You are working in a software development agency and a project manager and software architect approach you telling you that you're assigned to work on a new project. You are working on a {{ app_type }} called "{{ name }}" and you need to create a detailed development plan so that developers can start developing the app.
 
 {{ project_details }}
-
+{{ features_list }}
```
senko, line 4: What about adding `features_list` to project_details instead so it doesn't need to be added to every other prompt?

PR #676, opened 2024-02-22
Fix sending changes back to rework if nothing got applied.

senko:
Also, try to tone down LLMs se of INPUT_REQUIRED.

### DIFF
--- a/pilot/helpers/agents/CodeMonkey.py
+++ b/pilot/helpers/agents/CodeMonkey.py
@@ -8,6 +8,7 @@ from helpers.AgentConvo import AgentConvo
 from helpers.Agent import Agent
 from helpers.files import get_file_contents
 from const.function_calls import GET_FILE_TO_MODIFY, REVIEW_CHANGES
+from logger.logger import logger
 
 from utils.exit import trace_code_event
 from utils.telemetry import telemetry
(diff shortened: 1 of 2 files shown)


pilot/helpers/agents/CodeMonkey.py
```diff
@@ -268,17 +269,27 @@ def review_change(
...
             print("Applying entire change")
+            logger.info(f"Applying entire change to {file_name}")
             return new_content, None
+
         elif len(hunks_to_apply) == 0:
-            print(f"Rejecting entire change with reason: {llm_response['review_notes']}")
-            # If everything can be safely ignoring, it's probably because the files already implement the changes
-            # from previous tasks (which can happen often). Insisting on a change here is likely to cause problems.
-            return old_content, None
+            if hunks_to_rework:
```
senko, line 276: We're not always sending the `review_log` because it is always non-empty (because it includes `review_notes`). If we applied nothing and asked to rework, we need the log, otherwise if we applied nothing and ignored everything (asked nothing for rework), we don't want for the code monkey to rework it.

pilot/prompts/development/implement_changes.prompt
```diff
@@ -11,6 +11,6 @@ the full contents of the updated file, without skipping over any content
...
-**IMPORTANT**: If the user must configure something manually, mark the line that needs user configuration with `INPUT_REQUIRED {input_description}` comment,  where `input_description` is a description of what needs to be added here by the user. Use appropriate syntax for comments in the file you're saving (for example `// INPUT_REQUIRED {input_description}` in JavaScript). If the file type doesn't support comments (eg JSON), don't add any.
+**IMPORTANT**: Avoid hardcoding configuration values (eg. port, database url or server secret string) in code. Instead, use appropriate settings file (for example `.env`) for all user-configurable settings. Mark line(s) that need user configuration with `INPUT_REQUIRED {input_description}` comment,  where `input_description` is a description of what needs to be added in that line by the user. Use appropriate syntax for comments in the file you're saving (for example `# INPUT_REQUIRED {input_description}` in `.env`). If the file type doesn't support comments (eg JSON), don't add any.
```
senko, line 14: It is the hope of this author that this heartfelt plea will induce the LLM to be more considerate with respect to peppering `INPUT_REQUIRED` throughout the code and the resulting mental anguish that causes to the user.

PR #694, opened 2024-03-02
Misc fixes

senko:
Changes:

* add more files to known filenames and default ignores
* gracefully ignore file creation/removal OS errors
* handle user inputs larger than 64k
* show nice error message for TokenLimitError
* fix incorrect llm response stored in case of TokenLimitError
* gracefully handle edge case when saving template files
* improve agentic prints in CodeMonkey

Here's how unhandled TokenLimitError looks:

Screenshot from 2024-03-02 10-23-42

(this is in task breakdown, this is because within some parts of developer we're asking the user if they want to retry the task, but in task breakdown there's nothing to retry).

### DIFF
--- a/pilot/const/common.py
+++ b/pilot/const/common.py
@@ -29,6 +29,7 @@ DEFAULT_IGNORE_PATHS = [
     '.next',
     '.DS_Store',
     '__pycache__',
+    "site-packages",
     'node_modules',
     'package-lock.json',
     'venv',
--- a/pilot/helpers/AgentConvo.py
+++ b/pilot/helpers/AgentConvo.py
@@ -68,7 +68,7 @@ class AgentConvo:
                                                   function_calls=function_calls, prompt_data=prompt_data,
                                                   temperature=self.temperature)
         except TokenLimitError as e:
-            save_development_step(self.agent.project, prompt_path, prompt_data, self.messages, '', str(e))
+            save_development_step(self.agent.project, prompt_path, prompt_data, self.messages, {"text": ""}, str(e))
             raise e
 
         # TODO: move this code to Developer agent - https://github.com/Pythagora-io/gpt-pilot/issues/91#issuecomment-1751964079
--- a/pilot/helpers/Project.py
+++ b/pilot/helpers/Project.py
@@ -408,6 +408,7 @@ class Project:
             inputs_required = self.find_input_required_lines(data['content'])
             for line_number, line_content in inputs_required:
                 user_input = None
+                print('', type='verbose', category='human-intervention')
                 print(color_yellow_bold(f'Input required on line {line_number}:\n{line_content}') + '\n')
                 while user_input is None or user_input.lower() not in AFFIRMATIVE_ANSWERS + ['continue']:
                     print({'path': full_path, 'line': line_number}, type='openFile')
--- a/pilot/helpers/Project.py
+++ b/pilot/helpers/Project.py
@@ -475,7 +475,7 @@ class Project:
             # - /pilot -> /pilot/
             # - \pilot\server.js -> \pilot\server.js
             # - \pilot -> \pilot\
-            KNOWN_FILES = ["makefile", "dockerfile", "procfile", "readme", "license"]  # known exceptions that break the heuristic
+            KNOWN_FILES = ["makefile", "dockerfile", "procfile", "readme", "license", "podfile"]  # known exceptions that break the heuristic
             KNOWN_DIRS = []  # known exceptions that break the heuristic
             base = os.path.basename(path)
             if (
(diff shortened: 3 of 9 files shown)


pilot/const/common.py
```diff
@@ -29,6 +29,7 @@
     '.next',
     '.DS_Store',
     '__pycache__',
+    "site-packages",
```
senko, line 32: This fixes problems when people create a Python virtual environment within a project, irrespective of how they name the directory. There's at least a couple of reported case of this exact reason for TokenLimitError, including in telemetry and discord.

pilot/helpers/AgentConvo.py
```diff
@@ -68,7 +68,7 @@ def send_message(self, prompt_path=None, prompt_data=None, function_calls: Funct
                                                   function_calls=function_calls, prompt_data=prompt_data,
                                                   temperature=self.temperature)
         except TokenLimitError as e:
-            save_development_step(self.agent.project, prompt_path, prompt_data, self.messages, '', str(e))
+            save_development_step(self.agent.project, prompt_path, prompt_data, self.messages, {"text": ""}, str(e))
```
senko, line 71: This is the only place in the codebase where we're saving `llm_response` as a string instead of a dict.

After this crash, if the person tries to reload the project, they get:

```
TypeError: string indices must be integers, not 'str':
pilot/helpers/agents/Developer.py:132
pilot/helpers/agents/Developer.py:92
pilot/helpers/Project.py:204
```

Exact code number varies by version but it's always here:

https://github.com/Pythagora-io/gpt-pilot/blob/e31c67667dfad32a3c83a6425b86b122b488245e/pilot/helpers/agents/Developer.py#L132

pilot/helpers/Project.py
```diff
@@ -560,7 +561,10 @@ def restore_files(self, development_step_id):
 
         clear_directory(self.root_path, ignore=self.files)
         for file_snapshot in file_snapshots:
-            update_file(file_snapshot.file.full_path, file_snapshot.content, project=self)
+            try:
```
senko, line 564: This can happen if/when people delete, rename or move the project directory, this at least makes us not crash in that case.

pilot/helpers/files.py
```diff
@@ -145,10 +145,18 @@ def clear_directory(directory: str, ignore: Optional[list[str]] = None):
             if matcher.ignore(full_path):
                 continue
 
-            os.remove(full_path)
+            try:
```
senko, line 148: Also catches some weird edge cases. If we can't remove a file here, oh well, not such a big deal, we don't want to crash over it.

pilot/helpers/files.py
```diff
@@ -145,10 +145,18 @@ def clear_directory(directory: str, ignore: Optional[list[str]] = None):
...
         # Delete empty subdirectories not in ignore list
         for d in dirs:
-            subdir_path = os.path.join(dpath, d)
-            if not os.listdir(subdir_path):
-                os.rmdir(subdir_path)
+            try:
+                subdir_path = os.path.join(dpath, d)
+                if not os.listdir(subdir_path):
+                    os.rmdir(subdir_path)
+            except:  # noqa
```
senko, line 160: Same as above, we don't want to crash over a weird edge case.

pilot/helpers/ipc.py
```diff
@@ -28,8 +28,17 @@ def listen(self):
             return
 
         while True:
-            data = self.client.recv(65536)
-            message = json.loads(data)
+
+            data = b''
+            while True:
+                data = data + self.client.recv(512 * 1024)
```
senko, line 34: People do in fact send more than 64kb of input, probably when they copy/paste the output.

This reads up to 512kb in a single chunk, and then if json load fails, which should only ever happen if it's incomplete, this tries to read more. In no case is this worse than previously, since even if it gets stuck in a forever-loop in some weird case, the alternative here would just be to crash.

pilot/templates/__init__.py
```diff
@@ -75,7 +75,9 @@ def apply_project_template(
     print(color_green_bold(f"Applying project template {template['description']}...\n"))
     logger.info(f"Applying project template {template_name}...")
 
-    project.save_files_snapshot(project.checkpoints['last_development_step']['id'])
+    last_development_step = project.checkpoints.get('last_development_step')
```
senko, line 78: A weird case where this isn't available, let's not crash in that case.

pilot/helpers/Project.py
```diff
@@ -408,6 +408,7 @@ def save_file(self, data):
             inputs_required = self.find_input_required_lines(data['content'])
             for line_number, line_content in inputs_required:
                 user_input = None
+                print('', type='verbose', category='human-intervention')
                 print(color_yellow_bold(f'Input required on line {line_number}:\n{line_content}') + '\n')
```
LeonOstrez, line 412: instead of having extra print you can just do ` print(color_yellow_bold(f'Input required on line {line_number}:\n{line_content}') + '\n', type='verbose', category='human-intervention')`

pilot/helpers/agents/CodeMonkey.py
```diff
@@ -111,6 +113,7 @@ def implement_code_changes(
                 # No rework needed, we're done here
                 break
 
+            print('', type='verbose', category='agent:code-monkey')
```
LeonOstrez, line 116: only effect of this line is that it prints empty line, `category='agent:code-monkey'` is already set few lines above (line 111)

PR #699, opened 2024-03-04
report expired bricks keys to extension

senko:
Report expired bricks keys to extension


pilot/utils/llm_connection.py
```diff
@@ -440,6 +440,10 @@ def return_result(result_data, lines_printed):
         timeout=(API_CONNECT_TIMEOUT, API_READ_TIMEOUT),
     )
 
+    if response.status_code == 401 and 'BricksLLM' in response.text:
```
LeonOstrez, line 443: we should check what other status_codes and responses we can get from them

PR #702, opened 2024-03-05
multiple small fixes

LeonOstrez:


pilot/helpers/Project.py
```diff
@@ -233,11 +237,13 @@ def finish(self):
...
             if not self.skip_steps:
                 print('', type='verbose', category='pythagora')
+                if self.run_command and self.check_ipc():
+                    print(self.run_command, type='run_command')
                 feature_description = ask_user(self, "Project is finished! Do you want to add any features or changes? "
                                                      "If yes, describe it here and if no, just press ENTER",
                                                require_some_input=False)
 
-                if feature_description == '':
+                if feature_description == '' or feature_description == 'continue':
```
senko, line 246: can there ever be a "continue" button here?

LeonOstrez: not sure why but when there is `run_command` extension shows `Continue` button 🤷‍♂️  I guess they come in pair...

PR #758, opened 2024-03-19
github action using docker + cloudflare

abdulrahman305:
github action using docker + cloudflare


senko:
Thank you for the contribution, but currently we don't want to modify our CI/CD workflow in a way that would extend the infrastructure surface area and maintenance load.

PR #759, opened 2024-03-19
Experimental support for Anthropic Claude API

senko:
This adds support for Anthropic Claude via its new conversational API, using Anthropic Python SDK.

To enable the support, modify `.env` to add:
```
ANTHROPIC_API_KEY="sk-ant-api03-yourkey"
MODEL_NAME=anthropic/claude-3-opus-20240229
```

If proxying the Anthropic API through something else (must support its new conversational API), you can also set `ANTHROPIC_ENDPOINT`.

Sonnet and/or Haiku can also be used (Haiku probably won't work well for all the agents tho).

Also supports `--model-name` setting to override the model name at runtime (to be used by the Pythagora extension eventually).

## Tests

I added ANTHROPIC and OPENAI prints to distinguish between the two in the screenshots.

With this merged, default (empty) `.env`:

Screenshot from 2024-03-19 16-31-31

With this merged, `.env` with OpenAI keys:

Screenshot from 2024-03-19 16-33-08

With this merged and Anthropic keys in `.env`:

Screenshot from 2024-03-19 16-34-54

### DIFF
--- a/pilot/utils/llm_connection.py
+++ b/pilot/utils/llm_connection.py
@@ -109,8 +109,9 @@ def create_gpt_chat_completion(messages: List[dict], req_type, project,
              {'function_calls': {'name': str, arguments: {...}}}
     """
 
+    model_name = os.getenv('MODEL_NAME', 'gpt-4')
     gpt_data = {
-        'model': os.getenv('MODEL_NAME', 'gpt-4'),
+        'model': model_name,
         'n': 1,
         'temperature': temperature,
         'top_p': 1,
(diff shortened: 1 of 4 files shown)


pilot/utils/llm_connection.py
```diff
@@ -588,3 +599,48 @@ def postprocessing(gpt_response: str, req_type) -> str:
 
 def load_data_to_json(string):
     return json.loads(fix_json(string))
+
+
+
+def stream_anthropic(messages, function_call_message, gpt_data, model_name = "claude-3-sonnet-20240229"):
+    try:
+        import anthropic
```
senko, line 607: This is imported here so anyone updating (including through the extension) doesn't need to install the anthropic SDK unless they actually use this.

LeonOstrez (commented):
We should add to .env.example `ANTHROPIC_API_KEY`, `ANTHROPIC_ENDPOINT ` and `MODEL_NAME` with example how to set MODEL_NAME.

senko:
> We should add to .env.example `ANTHROPIC_API_KEY`, `ANTHROPIC_ENDPOINT ` and `MODEL_NAME` with example how to set MODEL_NAME.

Added in 71b8d77de77623243fbd93c9f778ee21345eb96f

PR #788, opened 2024-03-24
Loop

LeonOstrez:
- remove unused prompts
- add instructions to use relative paths in prompts
- fix storing of files in db (before we stored them twice, with and without starting "/")
- trigger get_run_command() only once per task


pilot/prompts/development/parse_task.prompt
```diff
@@ -22,15 +24,14 @@ Examples:
...
-        "path": "/server.js",
+        "path": "./server.js",
         "code_change_description": "",
       },
     },
     {
       "type": "command",
       "command": {
-        "command": "ls",
+        "command": "mv ./index.js ./public/index.js"",
```
senko, line 34: this is inconsistent with the expected success message and command id, could confuse the llm

pilot/helpers/agents/Developer.py
```diff
@@ -786,12 +791,13 @@ def review_code_changes(self):
         review_convo = AgentConvo(self)
         files = [
             file_dict for file_dict in self.project.get_all_coded_files()
-            if any(file_dict['full_path'].endswith(modified_file) for modified_file in self.modified_files)
+            if any(file_dict['full_path'].endswith(modified_file.lstrip('.')) for modified_file in self.modified_files)
```
senko, line 794: I don't understand this.

PR #814, opened 2024-03-29
Improve error handling with Anthropic API

ArtikUA:
The Anthropic API is limited for the new accounts when you have tier 1 or tier 2, so sometimes you hit it's limits.
Also, sometimes they API stopping working due to the limited resources of their serers

- Added @retry_on_exception for Anthropic
- Added `rate_limit_error` for the reasons to retry after the pause because of the rate limit exceeded to Anthropic
- Added `overloaded_error` for the reasons to retry after the pause because there's something wrong with Anthropic at the moment

### DIFF
--- a/pilot/utils/llm_connection.py
+++ b/pilot/utils/llm_connection.py
@@ -264,7 +264,12 @@ def retry_on_exception(func):
                     print(color_red(f"Error calling LLM API: The request exceeded the maximum token limit (request size: {n_tokens}) tokens."))
                     trace_token_limit_error(n_tokens, args[0]['messages'], err_str)
                     raise TokenLimitError(n_tokens, MAX_GPT_MODEL_TOKENS)
-                if "rate_limit_exceeded" in err_str:
+                if "rate_limit_exceeded" in err_str or "rate_limit_error" in err_str:
+                    # Retry the attempt if the current account's tier reaches the API limits
+                    rate_limit_exceeded_sleep(e, err_str)
+                    continue
+                if "overloaded_error" in err_str:
+                    # Retry the attempt if the Anthropic servers are overloaded
                     rate_limit_exceeded_sleep(e, err_str)
                     continue
 
@@ -601,7 +606,7 @@ def load_data_to_json(string):
     return json.loads(fix_json(string))
 
 
-
+@retry_on_exception
 def stream_anthropic(messages, function_call_message, gpt_data, model_name = "claude-3-sonnet-20240229"):
     try:
         import anthropic


senko:
Thanks @ArtikUA, this is great! :pray: 

Have you tested the rate limit error and overloaded error cases with your changes? Asking because `retry_on_exception` has a lot of built-in assumption about the underlying function (the default one) so it's not obvious if that's going to work well in the anthropic case.

Could you perhaps make it trigger this and then copy/paste or screenshot the output showing it in action?

ArtikUA:
@senko 
Yes, it's working, I reproduced it again just now:
image
image
Before that it was impossible to do anything because my app is quite bug and only Tier 2 in Anthropic

senko:
Great, thank you for the contribution! :pray:

senko:
@ArtikUA I think I'll have to wrapper decorator for anthropic, it reliably crashes for me if there's mismatch in the JSON output (eg. with Haiku not correctly following the output format):

```
Traceback (most recent call last):
  File "/home/senko/Projects/Pythagora/gpt-pilot/pilot/utils/llm_connection.py", line 216, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/senko/Projects/Pythagora/gpt-pilot/pilot/utils/llm_connection.py", line 649, in stream_anthropic
    assert_json_schema(response, gpt_data["functions"])
  File "/home/senko/Projects/Pythagora/gpt-pilot/pilot/utils/llm_connection.py", line 596, in assert_json_schema
    parsed = json.loads(response)
             ^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
               ^^^^^^^^^^^^^^^^^^^^^^
json.decoder.JSONDecodeError: Invalid control character at: line 3 column 1446 (char 1472)
```

This is because the wrapper function assumes the underlying function is called in a very special way, which is really just applicable to `stream_gpt_completion`.

It would be better to add retry mechanism directly in `stream_anthropic` (eg. a for loop that does a few retries on retryable errors). Would you be interested in hacking on that?

JamesD-git:
Screenshot 2024-04-23 at 10 19 30 AM
I've implemented this code but it's ended up in an infinite loop - I've confirmed in another app "test.py" that my t/min limit has reset

Any thoughts/advice?

PR #827, opened 2024-03-30
filter task-relevant files based on file summaries

senko:
This reduces the context we send for almost every LLM call, which improves focus on relevant files, increases speed, and reduces token usage and cost.

Gated behind a feature flag `FILTER_RELEVANT_FILES` which is disabled by default. To enable it, set `FILTER_RELEVANT_FILES=true` in `.env`.

Loading existing project will go through the files and retrieve summaries for them, after which they can be used to select only the relevant ones.

Tested on OpenAI (uses gpt3.5 for file summaries) and Anthropic (uses claude3 haiku for file summaries) LLMs.


pilot/helpers/AgentConvo.py
```diff
@@ -179,7 +179,8 @@ def load_branch(self, branch_name, reload_files=True):
             self.replace_files()
 
     def replace_files(self):
-        files = self.agent.project.get_all_coded_files()
+        relevant_files = getattr(self.agent, 'relevant_files', None)
```
senko, line 182: Keeping track of relevant files is per agent. If the agent doesn't keep track of files, there's no filtering.

Currently only Developer keeps track of relevant files. For example, CodeMonkey doesn't, but it's getting the AgentConvo from the Developer which does, so it also benefits from the filtering.

pilot/helpers/Project.py
```diff
@@ -342,8 +360,35 @@ def get_all_coded_files(self):
...
+        """
+        Get all coded files in the project.
+
+        Returns:
+            list: A list of coded files.
+        """
+        files = self.get_all_database_files()
+        if relevant_files:
+            n_files0 = len(files)
+            files = [file for file in files if self.relpath(file) in relevant_files]
```
senko, line 373: This is the actual filtering step. The path normalization (`self.relpath`) is there so we can match file path properly with the relevant files set.

pilot/helpers/Project.py
```diff
@@ -342,8 +360,35 @@ def get_all_coded_files(self):
...
+
         return self.get_files([file.path + '/' + file.name for file in files])
 
+    def get_file_summaries(self) -> Optional[dict[str, str]]:
+        """
+        Get summaries of all coded files in the project.
+
+        :returns: A dictionary of file summaries, or None if file filtering is not enabled.
+        """
+        if os.getenv('FILTER_RELEVANT_FILES', '').lower().strip() not in ['true', '1', 'yes', 'on']:
```
senko, line 386: If not enabled, returns `None` as the summaries, meaning filtering is disabled.

pilot/helpers/Project.py
```diff
@@ -405,11 +450,14 @@ def save_file(self, data):
 
         if path and path[0] == '/':
             path = path.lstrip('/')
-        (File.insert(app=self.app, path=path, name=name, full_path=full_path)
+
+        description = describe_file(self, relpath(abspath(full_path), abspath(self.root_path)), data['content'])
```
senko, line 454: Update description on every file save.

pilot/prompts/components/files_list.prompt
```diff
@@ -1,6 +1,8 @@
-{% if files|length > 0 %}Here are files that are currently implemented:
+{% if file_summaries %}These files are currently implemented:{% for fpath, summary in file_summaries.items() %}
+* `{{ fpath }}`: {{ summary }}{% endfor %}
+{% endif %}{% if files|length > 0 %}Here are the relevant files:
 ---START_OF_FILES---{% for file in files %}
-**{{ file.path }}/{{ file.name }}** ({{ file.lines_of_code }} lines of code):
+**{% if file.path %}{{ file.path }}/{% endif %}{{ file.name }}** ({{ file.lines_of_code }} lines of code):
```
senko, line 5: This avoids turning files in the project root (like `app.js`) into `/app.js` because the path is empty.

pilot/prompts/components/files_list.prompt
```diff
@@ -1,6 +1,8 @@
-{% if files|length > 0 %}Here are files that are currently implemented:
+{% if file_summaries %}These files are currently implemented:{% for fpath, summary in file_summaries.items() %}
```
senko, line 1: This is intentionally outside `START_OF_FILES` so the summaries don't get replaced.

pilot/prompts/development/review_task.prompt
```diff
@@ -13,16 +13,7 @@ Development process of this app was split into smaller tasks. Here is the list o
...
-{% if files|length > 0 %}---start_of_current_files---
-{% for file in files %}
-**{{ file.path }}/{{ file.name }}** ({{ file.lines_of_code }} lines of code):
-```
-{{ file.content }}
-```
-{% endfor %}
----end_of_current_files---
-{% endif -%}
+{{ files_list }}
```
senko, line 16: Since now we're including summaries + relevant files, we don't need special cased for current files.

pilot/templates/__init__.py
```diff
@@ -88,7 +88,7 @@ def apply_project_template(
 
     last_development_step = project.checkpoints.get('last_development_step')
     if last_development_step:
-        project.save_files_snapshot(last_development_step['id'])
+        project.save_files_snapshot(last_development_step['id'], summaries=template.get("files"))
```
senko, line 91: No need to ask LLMs to describe files we already have in a template, just get summaries from the template metadata.

pilot/utils/describe.py
```diff
@@ -0,0 +1,78 @@
...
+    else:
+        # Unknown default model (possibly local LLM), didsable file summaries
+        return ''
+
+    if not content or not content.strip():
+        return "(empty)"
+
+    logger.info("Calling LLM to summarize file %s", fpath)
+    try:
+        response_text = create_gpt_chat_completion(
```
senko, line 65: This is intentionally not using `AgentConvo` because it uses `replace_files` which saves file snapshots which calls this function...

PR #832, opened 2024-03-31
Improved code extraction for Claude

ArtikUA:
There's a problem with Claude that it adding some comments before or after the triple backticks, for example:

> Here is the updated version of the `submissionController.js` file with the requested changes:
> 
> \`\`\`
> const Submission = require('../models/Submission');

So the next changes extracts the code block from the given content by removing any surrounding text, triple backticks, language specifiers, and extra whitespace.
Then it's add an extra a single trailing newline to meet coding standards so IDE and git won't show any warnings.

I tested - it worked perfectly with Claude and OpenAI.
Before it spent around 1/3 of the time debugging and removing that text after it was saved into the files even with Opus.

### DIFF
--- a/pilot/helpers/agents/CodeMonkey.py
+++ b/pilot/helpers/agents/CodeMonkey.py
@@ -122,7 +122,7 @@ class CodeMonkey(Agent):
                 "rework_feedback": rework_feedback,
             })
             if content:
-                content = self.remove_backticks(content)
+                content = self.extract_code(content)
             convo.remove_last_x_messages(2)
 
         # If we have changes, update the file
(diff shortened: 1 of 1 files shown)


senko:
Thanks for the contribution @ArtikUA !

However CodeMonkey could also be used to output documentation or other files in which "```" may potentially appear in the middle of the file. In that case, it would mess up the output.

ArtikUA:
> Thanks for the contribution @ArtikUA !
> 
> However CodeMonkey could also be used to output documentation or other files in which "```" may potentially appear in the middle of the file. In that case, it would mess up the output.

Got it, I'll close this one, and maybe will come up with something better later.

PR #859, opened 2024-04-10
Update parse_task.prompt

MaxAnfilofyev:
rewrote for conciseness and clarity. Added examples to [command] to nudge LLM to give more tasks to the system and fewer tasks to a human


LeonOstrez:
Some of the changes seem too big for merge without detailed testing in eval tool. Also, there are a lot of major changes since v0.2 that affect this PR

PR #884, opened 2024-04-17
to simplify .env and make local llms work properly.

therealinvoker:
I attempted to simplify how .env interacts with llm_connection. LIteLLM wasn't working well due to extra keys. Openrouter wasn't working because provider needed to be part of the model name.


LeonOstrez:
with v0.2 we are not using `.env` anymore but `config.json` and we are handling LLM requests differently

PR #887, opened 2024-04-22
project complexity check

LeonOstrez:
Check project complexity and for simple projects force plan to have only 1 task to stop overengineering

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,11 +30,16 @@ jobs:
       matrix:
         # Test latest and oldest supported Python releases
         # See https://devguide.python.org/versions/
-        python-version: ['3.9', '3.12']
+        python-version: ['3.11', '3.12']
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:
     - uses: actions/checkout@v4
+
+    - name: Set up Node.js 20
+      uses: actions/setup-node@v3
+      with:
+        node-version: 20
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4
--- a/pilot/prompts/spec_writer/app_complexity.prompt
+++ b/pilot/prompts/spec_writer/app_complexity.prompt
@@ -1,0 +1,8 @@
+```
+{{ app_summary }}
+```
+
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -30,12 +30,17 @@ jobs:
...
         # Test latest and oldest supported Python releases
         # See https://devguide.python.org/versions/
-        python-version: ['3.9', '3.12']
+        python-version: ['3.11', '3.12']
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:
     - uses: actions/checkout@v4
 
+    - name: Set up Node.js 20
```
senko, line 39: when is this needed?

.github/workflows/ci.yml
```diff
@@ -30,12 +30,17 @@ jobs:
       matrix:
         # Test latest and oldest supported Python releases
         # See https://devguide.python.org/versions/
-        python-version: ['3.9', '3.12']
+        python-version: ['3.11', '3.12']
```
senko, line 33: I'd prefer we test for 3.9 while we still support 3.9 to avoid any subtle errors creeping up and bricking our users' installations for those that are on 3.9

pilot/utils/exit.py
```diff
@@ -100,10 +100,10 @@ def ask_user_email(project):
 
 def exit_gpt_pilot(project, ask_feedback=True):
     terminate_running_processes()
-    path_id = get_path_id()
+    # path_id = get_path_id()
```
senko, line 103: trust in git. delete instead of commenting out.

pilot/prompts/spec_writer/app_complexity.prompt
```diff
@@ -0,0 +1,8 @@
+```
+{{ app_summary }}
+```
+
+The following is a user prompt for application/software tool they are trying to develop. Determine the complexity of the user's request. Do NOT respond with thoughts, reasoning, explanations or anything similar, return ONLY a numerical representation of the complexity level. Use the following scale:
```
senko, line 5: OMG that's all there is to it?  :exploding_head: Nice!

PR #927, opened 2024-05-16
Add chinese support for promot.(In .env, let 'LANGUAGE=cn')

zkewind:
Add chinese support for promot.(In .env file, let 'LANGUAGE=cn')


senko:
Hi @zkewind,

since we've released 0.2, we won't be be merging older PRs to 0.1, but we've added an option to add additional directories for searching prompt files, by adding this to config.json:

```json
  "prompt": {
    "paths": [
      "/path/to/your/custom/prompts",
      "/path/to/gpt-pilot/core/prompts"
    ]
  }
```

Then, when the prompts are loaded, they will first be looked into the subdirectories of your custom prompts. If not found, the lookup will go to the default (second) directory. This means you can override those that you need, and this can be for a new language, or if you want to adapt them for a specific LLM, etc...

Although the structure and template tags are changed a bit, the prompts themselves are mostly the same, so hopefully it'll be asier to adapt the prompts to the new version.

PR #940, opened 2024-05-23
update prompts to latest version

LeonOstrez:
- updated prompts
- added getting of relevant files after task review and iteration also (they might expand requirements of the task)
- added schema for relevant files


.gitignore
```diff
@@ -5,6 +5,8 @@ __pycache__/
 htmlcov/
 dist/
 workspace/
+pilot/
```
senko, line 8: Why? we have `pilot/` in the new codebase (need it for the import to go smoothly, although we'll get rid of it).

.gitignore
```diff
@@ -16,3 +18,4 @@ config.json
 poetry.lock
 .DS_Store
 *.log
+brija.py
```
senko, line 21: nitpick: no need to add random garbage to .gitignore, you can just remove the files from your directory (or add to your personal ~/.gitignore)

core/agents/developer.py
```diff
@@ -197,18 +205,27 @@ async def breakdown_current_task(self) -> AgentResponse:
         self.set_next_steps(response, source)
         return AgentResponse.done(self)
 
-    async def get_relevant_files(self) -> AgentResponse:
+    async def get_relevant_files(
+            self,
+            user_feedback: Optional[str] = '',
```
senko, line 210: Nitpick: `Optional` means the type of the parameter can be `None`, not that the parameter is optional :D so this should just be: `user_feedback: str` (or better `Optional[str] = None`, which is what we use elsewhere)

core/agents/developer.py
```diff
@@ -197,18 +205,27 @@ async def breakdown_current_task(self) -> AgentResponse:
...
+        convo = AgentConvo(self).template(
+            "filter_files",
+            current_task=self.current_state.current_task,
+            user_feedback=user_feedback,
+            solution_description=solution_description,
+        ).require_schema(RelevantFiles)
 
         # FIXME: this doesn't validate correct structure format, we should use pydantic for that as well
-        llm_response: list[str] = await llm(convo, parser=JSONParser(), temperature=0)
+        llm_response: list[str] = await llm(convo, parser=JSONParser(RelevantFiles), temperature=0)
```
senko, line 225: Since we're not validating the format, you can remove the FIXME above it

core/prompts/partials/relative_paths.prompt
```diff
@@ -1 +1,4 @@
-**IMPORTANT**: Pay attention to file paths: if the command or argument is a file or folder from the project, use paths relative to the project root (for example, use `somedir/somefile` instead of `/path/to/project/somedir/somefile`).
+**IMPORTANT**: Pay attention to file paths: if the command or argument is a file or folder from the project, use paths relative to the project root.
+ For example:
+ - use `./dirname/filename.py` instead of `/path/to/project/dirname/filename.py`
```
senko, line 3: Without `./` in the new code, just `dirname/filename.py`.

example-config.json
```diff
@@ -17,7 +17,7 @@
   "agent": {
     "default": {
       "provider": "openai",
-      "model": "gpt-4-turbo",
+      "model": "gpt-4-turbo-preview",
```
senko, line 20: You probably want to change it in `AgentLLMConfig` as well.

senko (approved):
Looks mostly good, I added a few nitpick comments, the only one that I think is important is in relative paths (we don't need no `./` any longer), but the others are trivial to do as well.

PR #941, opened 2024-05-23
Prevent loops in task reviewer

gperetin:
I'm not sure this is correct, having a hard time with the verification.

I was able to get it to have a `review` source, and it doesn't get stuck, but I wasn't able to get a reproduction on the current state.

I also have some error entries that seem related, but I've had those before this change as well, which is interesting.

```
2024-05-23 14:35:53,981 WARNING [core.agents.developer] Unfinished steps found before the next iteration is broken down: [{'id': '53a7b3eb566a433bb1579eeb39565a38', 'completed': False, 'type': 'review_task', 'source': 'feature'}, {'id': 'f114d468eca54614b5bba715ebd6aaa9', 'completed': False, 'type': 'create_readme', 'source': 'feature'}]
```
README does get updated at the end of the task so I think that part is correct.

### DIFF
--- a/core/agents/developer.py
+++ b/core/agents/developer.py
@@ -100,12 +100,6 @@ class Developer(BaseAgent):
         :param review_feedback: If provided, the task review feedback is broken down instead of the current iteration
         :return: AgentResponse.done(self) when the breakdown is done
         """
-        if self.current_state.unfinished_steps:
-            # if this happens, it's most probably a bug as we should have gone through all the
-            # steps before getting new new iteration instructions
-            log.warning(
-                f"Unfinished steps found before the next iteration is broken down: {self.current_state.unfinished_steps}"
-            )
 
         if review_feedback is not None:
             iteration = None
--- a/core/agents/developer.py
+++ b/core/agents/developer.py
@@ -241,7 +241,7 @@ class Developer(BaseAgent):
             }
             for step in response.steps
         ]
-        if len(self.next_state.unfinished_steps) > 0:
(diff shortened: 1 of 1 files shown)


senko:
> Unfinished steps found before the next iteration is broken down

This happens because earlier the only time iteration breakdown function should be called is after all the steps are done and then Troubleshooter is run. Then the Troubleshooter asks the user to test the app and if there's a problem, creates an "iteration" which is an attempt at fixing the problem.

However, now this function is also run because TaskReviewer asks Developer to do so (via an AgentResponse). At the time when The Task Reviewer does so, there are two remaining steps:
* task review (this is what Task Reviewer is currently doing)
* create readme

These are the two steps visible in the warning. Now,  the important bit here is that if one agent calls another (or rather, tells Orchestrator it needs something, and then Orchestrator calls the aother agent), the current step is not committed. So even though TaskReviewer marked its own step as complete, that's till in the "next state" by the time the iteration breakdown function is ran.

The solution is simply to remove the warning. It's now obsolete :)

LeonOstrez:
@gperetin  you can remove the warning then and this is ready for merge. Did you manage to test this works as intended?

gperetin:
> @gperetin you can remove the warning then and this is ready for merge. Did you manage to test this works as intended?

I have not, got stuck on that missing prompt template issue. I just sent out a PR for the LLM check, so getting back on this now.

gperetin:
I was able to test this and works as expected, meaning, after the troubleshooter step, README was not updated.

PR #949, opened 2024-05-24
Check LLM API before doing anything else.

gperetin:
Moved this out of `Orchestrator` so it can be done before we even have an agent. Also, didn't like the previuos logic in the `Orchestrator`, it was only there becuase we had an access to an LLM there.

Fail if a single check fails.

Tested by fiddling with config, here's what it looks like if it fails:
image

```
❯ python main.py --local-ipc-port 8125 --project aa74d7405b3144ae9077ac2af04a816e
2024-05-24 11:11:24,731 WARNING [core.cli.main] API check for openai failed with: The model `gpt-4-turbo123` does not exist or you do not have access to it.
```

Only tests once per LLM provider.

@senko if you could take a peek at that telemetry entry, I see `set` entries in the `async_main` but I don't see a start there, am I doing something wrong?

### DIFF
--- a/core/cli/main.py
+++ b/core/cli/main.py
@@ -115,6 +115,36 @@ async def start_new_project(sm: StateManager, ui: UIBase) -> bool:
     return project_state is not None
 
 
+async def run_pythagora_session(sm: StateManager, ui: UIBase, args: Namespace):
+    """
+    Run a Pythagora session.
+
+    :param sm: State manager.
+    :param ui: User interface.
+    :param args: Command-line arguments.
+    :return: True if the application ran successfully, False otherwise.
+    """
+
+    if not await llm_api_check(ui):
+        return False
+
+    if args.project or args.branch or args.step:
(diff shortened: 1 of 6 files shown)


core/cli/main.py
```diff
@@ -112,6 +153,9 @@ async def async_main(
     if not ui_started:
         return False
 
+    if not await llm_api_check(ui):
```
senko, line 156: I previously initialized `telemetry.start()` and then did `telemetry.send()` in `run_project()`. Because it is now needed before that, I think it makes sense to do `start()` before this line, but we don't want to run it for list/config/import/... operations, I think it makes sense to split this function into two:

1. async_main() that runs everything before this line, and then runs the second one
2. run_pythagora_session() that runs everything from this line down

Then the async main can have telemetry start, run the new function, telemetry send.

(and the start and send can be removed from run_project). This will minimize duplication and still make it start/end correctly

Lemme know if that makes sense

PR #953, opened 2024-05-24
Implement missing VSCode IPC messages

senko:
Implement missing VSCode IPC messages:

* loadingFinished
* projectDescription
* featuresList

+ minor prompt and button fixes

### DIFF
--- a/core/agents/tech_lead.py
+++ b/core/agents/tech_lead.py
@@ -134,6 +134,7 @@ class TechLead(BaseAgent):
         self.next_state.tasks = self.current_state.tasks + [
             {
                 "id": uuid4().hex,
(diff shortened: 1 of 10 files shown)


core/agents/tech_lead.py
```diff
@@ -134,6 +134,7 @@ async def plan_epic(self, epic) -> AgentResponse:
         self.next_state.tasks = self.current_state.tasks + [
             {
                 "id": uuid4().hex,
+                "epic_id": epic["id"],
```
LeonOstrez, line 137: we can remove this since we agreed not to save all tasks in the state but only tasks from the current epic

PR #955, opened 2024-05-25
add action label to states for easier debugging/loading

senko:
This adds a new `action` field to ProjectState, which makes it easy to see what's going on in each step. This is useful both when loading the project and when inspecting the database.

### Project load

Screenshot from 2024-05-25 11-01-49

(note: dropdown height tweked just for the demo, so more items can be seen)

### Database inspection

From the `project_state` table:

Screenshot from 2024-05-25 11-10-42


core/agents/developer.py
```diff
@@ -153,6 +153,10 @@ async def breakdown_current_iteration(self, review_feedback: Optional[str] = Non
 
         if iteration:
             self.next_state.complete_iteration()
+            self.next_state.action = f"Troubleshooting #{len(self.current_state.iterations)}"
+        else:
+            self.next_state.complete_task()
```
LeonOstrez, line 158: why do we complete the task if the reviewer added steps?

PR #957, opened 2024-05-25
Fixes

LeonOstrez:
Few smaller fixes:

- sending user instructions for testing to extension as a hint
- cleanup tasks after each epic. Currently, feature tasks are just added to list of app tasks.
- fixed sending task description in question
- added logs of example project description and example task description
- removed "skip task" button if there is only 1 task in epic (we cant skip only task)

### DIFF
--- a/core/agents/developer.py
+++ b/core/agents/developer.py
@@ -262,10 +262,16 @@ class Developer(BaseAgent):
 
         :return: True if the task should be executed as is, False if the task is skipped or edited
         """
+        buttons = {"yes": "Yes", "edit": "Edit Task"}
+        if len(self.current_state.tasks) > 1:
+            buttons["skip"] = "Skip Task"
+
         description = self.current_state.current_task["description"]
+        await self.send_message("Starting new task with description:")
+        await self.send_message(description)
         user_response = await self.ask_question(
-            "Do you want to execute this task?",
-            buttons={"yes": "Yes", "edit": "Edit Task", "skip": "Skip Task"},
+            "Do you want to execute the above task?",
+            buttons=buttons,
             default="yes",
             buttons_only=True,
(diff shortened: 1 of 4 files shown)


core/db/models/project_state.py
```diff
@@ -249,6 +249,7 @@ def complete_epic(self):
 
         log.debug(f"Completing epic {self.unfinished_epics[0]['name']}")
         self.unfinished_epics[0]["completed"] = True
+        self.tasks = []
```
senko, line 252: Hmmm I was thinking that we might save tasks (so I started adding epic_id to each of them so we can easily find them). I think we might need this info for later prompts. 

If this is just to support sending the list of tasks to vscode extension just for current extension, we should be able to filter them on epic_id (which may not exist as in theory we've talked about tasks not belonging to any epic in some special cases). So the decision to clear or not should be based on whether we'd need them in future prompts (and we always want to avoid reading previous states from pythagora).

core/agents/developer.py
```diff
@@ -262,12 +262,19 @@ async def ask_to_execute_task(self) -> bool:
 
         :return: True if the task should be executed as is, False if the task is skipped or edited
         """
+        buttons = {"yes": "Yes", "edit": "Edit Task"}
+        if len(self.current_state.tasks) > 1:
+            buttons["skip"] = "Skip Task"
```
senko, line 267: We intentionally don't want to allow skipping the first task?

LeonOstrez: this will show the "Skip task" button to the user only if Epic has more than 1 task. If there is only 1 task in epic we dont want to allow user to skip because then epic would be marked as finished/done and nothing was implemented. Makes no sense to skip a task if there is only one task in the whole epic

PR #960, opened 2024-05-25
Prevent lazy loading of SQLAlchemy relationships.

gperetin:
I had an issue with one relationship, which is `FileContent.files`. When I added `lazy="raise"` there, `import_files()` in `StateManager` started failing, and I had to add an `await` there. I'm not sure why that fixes it.

I'm not sure I understand what "selectin" does, accorting to the docs (https://docs.sqlalchemy.org/en/20/orm/relationship_api.html), "selectin" is an eagerly loading strategy, so not sure why we have to do `await obh.awaitable_attrs.field`.


senko (approved):
Seems legit! :shipit:

PR #966, opened 2024-05-27
add TaskCompleter Agent

LeonOstrez:
- added TaskCompleter Agent and moved complete_task() from all other Agents to new one
- removed task["completed"] boolean and replaced with task["status"] which can be in multiple stages
- fixed issue where Readme was created after every iteration, now it is not a "step" but we have new flow after user reviews tasks and confirm everything works. Flow is TechnicalWriter -> TechLead (update epic) -> TaskCompleter

### DIFF
--- a/core/agents/developer.py
+++ b/core/agents/developer.py
@@ -248,12 +248,6 @@ class Developer(BaseAgent):
                     "type": "review_task",
                     "source": source,
                 },
-                {
-                    "id": uuid4().hex,
-                    "completed": False,
-                    "type": "create_readme",
-                    "source": source,
-                },
             ]
         log.debug(f"Next steps: {self.next_state.unfinished_steps}")
 
@@ -287,8 +281,8 @@ class Developer(BaseAgent):
         if user_response.cancelled or user_response.button == "skip":
             log.info(f"Skipping task: {description}")
             self.next_state.current_task["instructions"] = "(skipped on user request)"
-            self.next_state.complete_task()
-            await self.send_message(f"Skipping task {description}")
+            self.next_state.current_task["status"] = "skipped"
+            await self.send_message("Skipping task...")
             # We're done here, and will pick up the next task (if any) on the next run
             return False
 
--- a/core/agents/developer.py
+++ b/core/agents/developer.py
@@ -7,6 +7,7 @@ from pydantic import BaseModel, Field
 from core.agents.base import BaseAgent
 from core.agents.convo import AgentConvo
 from core.agents.response import AgentResponse, ResponseType
+from core.db.models.project_state import TaskStatus
 from core.llm.parser import JSONParser
 from core.log import get_logger
 
@@ -281,7 +282,7 @@ class Developer(BaseAgent):
         if user_response.cancelled or user_response.button == "skip":
             log.info(f"Skipping task: {description}")
             self.next_state.current_task["instructions"] = "(skipped on user request)"
-            self.next_state.current_task["status"] = "skipped"
+            self.next_state.set_current_task_status(TaskStatus.SKIPPED)
             await self.send_message("Skipping task...")
             # We're done here, and will pick up the next task (if any) on the next run
             return False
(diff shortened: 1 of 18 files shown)


core/agents/developer.py
```diff
@@ -287,8 +281,8 @@ async def ask_to_execute_task(self) -> bool:
         if user_response.cancelled or user_response.button == "skip":
             log.info(f"Skipping task: {description}")
             self.next_state.current_task["instructions"] = "(skipped on user request)"
-            self.next_state.complete_task()
-            await self.send_message(f"Skipping task {description}")
+            self.next_state.current_task["status"] = "skipped"
```
senko, line 284: This is a modification of a structure (dictionary within a list) inside the SQLALchemy ORM field. SQLalchemy won't detec it so it's very likely it won't make an update in the database.

For this reason, we have [`ProjectState.flag_tasks_as_modified()`](https://github.com/Pythagora-io/gpt-pilot/blob/49cb71121c6944934d04ea9a904ead974c287389/core/db/models/project_state.py#L282), to tell SQLalchemy the field should be updated.

And since the task status is being set in several places, I'd suggest creating `set_current_task_status()` helper method in ProjectState which would set it and also call flag modified so nobody forgets it somewhere.

core/agents/orchestrator.py
```diff
@@ -163,11 +164,23 @@ def create_agent(self, prev_response: Optional[AgentResponse]) -> BaseAgent:
             if prev_response.type == ResponseType.INPUT_REQUIRED:
                 # FIXME: HumanInput should be on the whole time and intercept chat/interrupt
                 return HumanInput(self.state_manager, self.ui, prev_response=prev_response)
-            if prev_response.type == ResponseType.UPDATE_EPIC:
-                return TechLead(self.state_manager, self.ui, prev_response=prev_response)
             if prev_response.type == ResponseType.TASK_REVIEW_FEEDBACK:
                 return Developer(self.state_manager, self.ui, prev_response=prev_response)
 
+        status = state.current_task.get("status") if state.current_task else None
```
senko, line 170: Shouldn't this go down between "Ask the developer to break down current task" and "Execute next step in the task" ? 

Also nitpick: I like the `current_task_status` (that you used in TechLead) name more, since it's more descriptive.

core/agents/task_completer.py
```diff
@@ -0,0 +1,23 @@
...
+
+log = get_logger(__name__)
+
+
+class TaskCompleter(BaseAgent):
+    agent_type = "pythagora"
+    display_name = "Pythagora"
+
+    async def run(self) -> AgentResponse:
+        self.next_state.complete_task()
```
senko, line 13: Suggestion: set state action string to "Task #X complete"

core/agents/orchestrator.py
```diff
@@ -163,11 +164,23 @@ def create_agent(self, prev_response: Optional[AgentResponse]) -> BaseAgent:
...
+        if status:
+            # Status of the current task is set first time after the task was reviewed by user
+            log.info(f"Status of current task: {status}")
+            if status == "reviewed":
```
senko, line 174: Can we have all the statuses listed TaskStatus.REVIEWED ... (etc), (same like we did with specification.Complexity?) Helps with us poor typo-friendly devs :D  No need to subclass Enum (since it complicates things a bit, hence I didn't do that when converting Complexity), but just having those as constants will make me sleep easier at night.

senko (commented):
Mostly makes sense to me (I know it's still WIP). Added comments about flagging orm fields as modified as it may not work correctly without it.

In general, what do we think about removing "completed" boolean now that we have the same but more in "status" ?

core/agents/orchestrator.py
LeonOstrez: haha anything for you, will change it

core/agents/task_completer.py
LeonOstrez: 👍

core/agents/orchestrator.py
LeonOstrez: I think i started implementation by placing it there but then moved it above. Don't remember why now, I'll take a look again.

PR #973, opened 2024-05-28
Retry when we get a generic LLM API error.

gperetin:
Ask the user to retry when we get `{LLM Provider}.APIerror` exception.

### DIFF
--- a/core/agents/base.py
+++ b/core/agents/base.py
@@ -135,7 +135,15 @@ class BaseAgent:
             )
             if answer.button == "yes":
                 return True
-
+        elif error == LLMError.GENERIC_API_ERROR:
+            await self.stream_handler(message)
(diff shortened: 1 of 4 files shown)


senko:
Reworked the PR so we alway auto-retry up to max retries (3). If that fails, we ask the user if they want to retry again, in which case we again retry up to 3 times, then ask again, etc. If the user declines, we raise APIError which stops Pythagora:

Screenshot from 2024-05-28 20-02-27

(between each of the "do you want to retry" there's 3 retries underneath)

gperetin:
LGTM, can't merge as it still has me as an author, :ship:

senko:
@LeonOstrez since Goran and me collaborated on this, it falls to you as an independent party to review :)

PR #976, opened 2024-05-29
support existing projects

senko:
Adds **EXPERIMENTAL** support for importing existing projects.

### Demo

https://github.com/Pythagora-io/gpt-pilot/assets/3362/882c12e7-2b7c-4676-b441-8651688f9bb9

### DIFF
--- a/core/agents/importer.py
+++ b/core/agents/importer.py
@@ -6,19 +6,11 @@ from core.agents.response import AgentResponse, ResponseType
 from core.db.models import Complexity
 from core.llm.parser import JSONParser
 from core.log import get_logger
-from core.templates.example_project import (
-    EXAMPLE_PROJECT_DESCRIPTION,
-)
-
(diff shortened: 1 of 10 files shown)


core/agents/importer.py
```diff
@@ -0,0 +1,86 @@
...
+        self.send_message("Analyzing project ...")
+
+        convo = AgentConvo(self).template(
+            "analyze_project", relevant_files=relevant_files, example_spec=EXAMPLE_PROJECT_DESCRIPTION
+        )
+        llm_response = await llm(convo)
+
+        spec = self.current_state.specification.clone()
+        spec.description = llm_response
+        self.next_state.specification = spec
```
LeonOstrez, line 73: Should we maybe also get architecture, sys dependencies and package dependencies?

LeonOstrez (commented):
In general I think it is good that with new agent we also create system message for that agent but if this works ok, we can skip it now

PR #1013, opened 2024-06-12
Code monkey in iteration

senko:
CodeMonkey prompt references a previous message which is either task breakdown or iteration, but we've previously always appended the task convo. Now, if a file is changed based on the iteration instructions, the correct prompt template is used

Prompts in LLM request log in database:

Screenshot from 2024-06-12 14-46-30

Also, a tiny fix for yet another variant of API error that we want to catch and handle.

### DIFF
--- a/core/agents/code_monkey.py
+++ b/core/agents/code_monkey.py
@@ -35,19 +35,29 @@ class CodeMonkey(BaseAgent):
         # FIXME: Current prompts reuse task breakdown / iteration messages so we have to resort to this
         task = self.current_state.current_task
         current_task_index = self.current_state.tasks.index(task)
+        convo = AgentConvo(self)
 
-        convo = AgentConvo(self).template(
-            "breakdown",
-            task=task,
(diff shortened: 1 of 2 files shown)


core/agents/code_monkey.py
```diff
@@ -35,19 +35,29 @@ def _get_task_convo(self) -> AgentConvo:
...
+                docs=self.current_state.docs,
+            ).assistant(iteration["description"])
         else:
-            convo.assistant(self.current_state.current_task["instructions"])
+            # Implementing file according to the task instructions
+            convo.template(
+                "breakdown",
+                task=task,
+                iteration=None,
+                current_task_index=current_task_index,
```
LeonOstrez, line 58: shouldn't we have `docs` also in the breakdown?

LeonOstrez:
since https://github.com/Pythagora-io/gpt-pilot/pull/1018 was merged, I'm closing this PR

PR #1014, opened 2024-06-12
Only add docs for non-simple epics.

gperetin:
Also, tell the LLM to limit the number of candidate docsets it suggests.

### DIFF
--- a/core/agents/orchestrator.py
+++ b/core/agents/orchestrator.py
@@ -19,6 +19,7 @@ from core.agents.tech_lead import TechLead
 from core.agents.tech_writer import TechnicalWriter
 from core.agents.troubleshooter import Troubleshooter
 from core.db.models.project_state import TaskStatus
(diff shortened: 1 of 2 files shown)


core/agents/orchestrator.py
```diff
@@ -198,7 +199,7 @@ def create_agent(self, prev_response: Optional[AgentResponse]) -> BaseAgent:
             # Ask the Tech Lead to break down the initial project or feature into tasks and apply project template
             return TechLead(self.state_manager, self.ui, process_manager=self.process_manager)
 
-        if state.current_task and state.docs is None:
+        if state.current_task and state.docs is None and state.specification.complexity != Complexity.SIMPLE:
```
LeonOstrez, line 202: this means we will not update `state.docs` so it will stay `None` which is different from `[]` which we have currently. Is that ok?

PR #1024, opened 2024-06-18
Add route files content to prompt for user instructions.

gperetin:
This fixes the issue where Pythagora would provide incorrect links in user testing instructions, because that's what we got from the LLM. LLM couldn't have know the correct routes because we never told it what those were, so this PR addressed that by adding the contents of the routes files into the prompt for user testing instructions.

Tested by loading 2 user databases that had the issue with incorrect links in user testing instructions, and confirming that the correct links are generated.

Went with the `gpt-4o` here as it was significantly more accurate on selecting files from a list.

### DIFF
--- a/core/prompts/troubleshooter/get_route_files.prompt
+++ b/core/prompts/troubleshooter/get_route_files.prompt
@@ -1,0 +1,22 @@
+{# eval-tool test ID: 74918173-59e4-4005-bf19-d28f3bc9f06c
+
+This was added in June 2024, to improve the accuracy of user testing instructions.
+We've noticed that the LLM would often times include incorrect URLs in the user testing
+instructions, because it wasn't aware of the routes used in the application.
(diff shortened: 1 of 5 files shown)


core/prompts/troubleshooter/get_route_files.prompt
```diff
@@ -0,0 +1,22 @@
...
+```
+{
+    "files": [
+        "dir/file1.js",
+        "dir/file2.js",
+    ]
+}
+```
+
+Your response must be a valid JSON format.
```
LeonOstrez, line 22: How come this is also needed if we are using json schema?

gperetin: I saw this in other agents (executor, architect, code monkey) that also use schema but specify that in the prompt, so I thought it was a best practice. Will remove.

invisiblepancake (commented):
=)

PR #1057, opened 2024-07-24
Implemented the first step in deep focus initiative. Now, when the us…

zvone187:
…er says there is a bug during the testing step, GPT Pilot adds logs before attempting to fix the issue


core/config/magic_words.py
LeonOstrez, line 1: 🤦

core/agents/troubleshooter.py
```diff
@@ -33,7 +33,29 @@ class Troubleshooter(IterationPromptMixin, BaseAgent):
     agent_type = "troubleshooter"
     display_name = "Troubleshooter"
 
-    async def run(self) -> AgentResponse:
+    async def run(self):
```
LeonOstrez, line 36: `run()` should return AgentResponse, when is that not the case? From what I see, we do return AgentResponse always but should double check

core/agents/problem_solver.py
```diff
@@ -98,6 +99,7 @@ async def try_alternative_solutions(self) -> AgentResponse:
         self.next_state_iteration["alternative_solutions"][index]["tried"] = True
         self.next_state_iteration["description"] = llm_solution
         self.next_state_iteration["attempts"] = self.iteration["attempts"] + 1
+        self.next_state_iteration["status"] = IterationStatus.PROBLEM_SOLVER
```
LeonOstrez, line 102: why is Problem Solver setting status `PROBLEM_SOLVER`? Can you double check this? It seems like Troubleshooter sets it and here it should be something else

core/agents/orchestrator.py
```diff
@@ -226,12 +227,23 @@ def create_agent(self, prev_response: Optional[AgentResponse]) -> BaseAgent:
             return self.create_agent_for_step(state.current_step)
 
         if state.unfinished_iterations:
-            if state.current_iteration["description"]:
-                # Break down the next iteration into steps
+            if state.current_iteration["status"] == IterationStatus.HUNTING_FOR_BUG:
```
LeonOstrez, line 230: it would help with readability, and to make coding easier in future, if you:
```
current_iteration_status = state.current_iteration["status"]
```
and then use that variable

core/agents/orchestrator.py
```diff
@@ -226,12 +227,23 @@ def create_agent(self, prev_response: Optional[AgentResponse]) -> BaseAgent:
...
+                # Ask the Logger to check if more logs in the code are needed
+                return BugHunter(self.state_manager, self.ui)
+            elif (state.current_iteration["status"] == IterationStatus.AWAITING_LOGGING or
```
LeonOstrez, line 233: maybe it is nicer like this:
```
elif current_iteration_status  in (
    IterationStatus.AWAITING_LOGGING,
    IterationStatus.AWAITING_BUG_FIX,
    IterationStatus.IMPLEMENT_SOLUTION
)

Comment (LeonOstrez) [reply core/agents/orchestrator.py:233]: Also please add comments why is Developer called for each case. It is also ok to do:
```
elif current_iteration_status === IterationStatus.AWAITING_LOGGING
    # WHY?
    return Developer(self.state_manager, self.ui)
elif current_iteration_status === IterationStatus. AWAITING_BUG_FIX
    # WHY?
    return Developer(self.state_manager, self.ui)
elif current_iteration_status === IterationStatus. IMPLEMENT_SOLUTION
    # WHY?
    return Developer(self.state_manager, self.ui)
```

File: core/agents/orchestrator.py [hunk 5]
```diff
@@ -226,12 +227,23 @@ def create_agent(self, prev_response: Optional[AgentResponse]) -> BaseAgent:
...
+                  state.current_iteration["status"] == IterationStatus.AWAITING_BUG_FIX or
+                  state.current_iteration["status"] == IterationStatus.IMPLEMENT_SOLUTION):
+                # Ask the Logger to ask user to test new logs
                 return Developer(self.state_manager, self.ui)
-            else:
-                # We need to iterate over the current task but there's no solution, as Pythagora
-                # is stuck in a loop, and ProblemSolver needs to find alternative solutions.
+            elif (state.current_iteration["status"] == IterationStatus.AWAITING_USER_TEST or
```
LeonOstrez, line 238: same as above, dont reuse comment `# Ask the Logger to ask user to test new logs`

core/agents/bug_hunter.py
```diff
@@ -0,0 +1,190 @@
...
+    ]
+
+
+class HuntConclusionType(str, Enum):
+    ADD_LOGS = magic_words.ADD_LOGS
+    PROBLEM_IDENTIFIED = magic_words.PROBLEM_IDENTIFIED
+
+
+class HuntConclusionOptions(BaseModel):
+    conclusion: HuntConclusionType = Field(description=f"If more logs are needed to identify the problem, respond with '{magic_words.ADD_LOGS}'. If the problem is identified, respond with '{magic_words.PROBLEM_IDENTIFIED}'.")
```
LeonOstrez, line 58: In my opinion, this is a bad implementation because we are relying on LLM following description which will in most cases (non OpenAI models) not work and we are blocking ourselves from potentially giving this to another LLM.

It might be better to put Enum for those 2 options in schema

core/agents/bug_hunter.py
```diff
@@ -0,0 +1,190 @@
...
+
+class StepType(str, Enum):
+    ADD_LOG = "add_log"
+    EXPLAIN_PROBLEM = "explain_problem"
+    GET_ADDITIONAL_FILES = "get_additional_files"
+
+
+class Log(BaseModel):
+    filePath: str
+    referenceCodeSnippet: str = Field(description="Five lines of code before the line where the log needs to be added. Make sure that this contains **ONLY** the code that is currently written in the file. It must not contain the log that you want to add.")
```
LeonOstrez, line 25: Is this description correct?

core/agents/developer.py
```diff
@@ -156,8 +126,14 @@ async def breakdown_current_iteration(self, task_review_feedback: Optional[str]
         self.set_next_steps(response, source)
 
         if iteration:
-            self.next_state.complete_iteration()
-            self.next_state.action = f"Troubleshooting #{len(self.current_state.iterations)}"
+            # fixme please :cry:
```
LeonOstrez, line 129: 🤔

core/agents/developer.py
```diff
@@ -156,8 +126,14 @@ async def breakdown_current_iteration(self, task_review_feedback: Optional[str]
...
+            if ("status" in iteration) and (iteration["status"] == IterationStatus.AWAITING_BUG_FIX or
```
LeonOstrez, line 130: This is not too readable and needs some comments for other developers 😄 I would suggest something like this:
```
            if "status" not in iteration:
                # This is just a support for old iterations that don't have status
                self.next_state.complete_iteration()
                self.next_state.action = f"Troubleshooting #{len(self.current_state.iterations)}"
            elif iteration["status"] == IterationStatus.AWAITING_BUG_FIX:
                # If bug fixing is done, ask user to test again
                self.next_state.current_iteration["status"] = IterationStatus.AWAITING_USER_TEST
            elif iteration["status"] == IterationStatus.AWAITING_LOGGING:
                # If logging is done, ask user to reproduce the bug
                self.next_state.current_iteration["status"] = IterationStatus.AWAITING_BUG_REPRODUCTION
```

core/agents/developer.py
```diff
@@ -109,6 +68,17 @@ async def breakdown_current_iteration(self, task_review_feedback: Optional[str]
             n_tasks = 1
             log.debug(f"Breaking down the task review feedback {task_review_feedback}")
             await self.send_message("Breaking down the task review feedback...")
+        elif (self.current_state.current_iteration["status"] == IterationStatus.AWAITING_BUG_FIX or
```
LeonOstrez, line 71: 🤮

core/agents/bug_hunter.py
```diff
@@ -0,0 +1,190 @@
...
+class HuntConclusionType(str, Enum):
+    ADD_LOGS = magic_words.ADD_LOGS
+    PROBLEM_IDENTIFIED = magic_words.PROBLEM_IDENTIFIED
+
+
+class HuntConclusionOptions(BaseModel):
+    conclusion: HuntConclusionType = Field(description=f"If more logs are needed to identify the problem, respond with '{magic_words.ADD_LOGS}'. If the problem is identified, respond with '{magic_words.PROBLEM_IDENTIFIED}'.")
+
+
+class BugHunter(BaseAgent):
```
LeonOstrez, line 61: We might need to ask Aashan to support this agent in extension, double check that

core/agents/bug_hunter.py
```diff
@@ -0,0 +1,190 @@
...
+
+        if "bug_reproduction_description" not in current_iteration:
+            await self.get_bug_reproduction_instructions()
+        if current_iteration["status"] == IterationStatus.HUNTING_FOR_BUG:
+            # TODO determine how to find a bug (eg. check in db, ask user a question, etc.)
+            return await self.check_logs()
+        elif current_iteration["status"] == IterationStatus.AWAITING_USER_TEST:
+            return await self.ask_user_to_test()
+        elif current_iteration["status"] == IterationStatus.AWAITING_BUG_REPRODUCTION:
+            return await self.ask_user_to_test()
```
LeonOstrez, line 76: can we please not do elif and elif and in both of them call same function without any difference? We can at least send some variable to function so that inside function we dont have to check iteration status

core/agents/bug_hunter.py
```diff
@@ -0,0 +1,190 @@
...
+            # if logs are needed, add logging steps
+            self.next_state.current_iteration["status"] = IterationStatus.AWAITING_LOGGING
+            await self.send_message("Adding more logs to identify the bug.")
+
+        self.next_state.flag_iterations_as_modified()
+        return AgentResponse.done(self)
+
+    async def ask_user_to_test(self):
+
+        reproduce_bug_and_get_logs = self.current_state.current_iteration["status"] == IterationStatus.AWAITING_BUG_REPRODUCTION
```
LeonOstrez, line 151: This is basically variable that i mentioned to pass in to the function so we dont check status here

core/agents/bug_hunter.py
```diff
@@ -0,0 +1,190 @@
...
+
+        await self.send_message("You can reproduce the bug like this:\n\n" + self.current_state.current_iteration["bug_reproduction_description"])
+        if self.current_state.current_iteration["status"] == IterationStatus.AWAITING_USER_TEST:
```
LeonOstrez, line 154: same here

LeonOstrez (changes requested):
Please check tests, they are failing

core/prompts/bug-hunter/system.prompt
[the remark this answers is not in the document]
zvone187: tbh, I don't know what to put into it so it's meaningful - I have a feeling it doesn't make any difference

core/prompts/bug-hunter/parse_task.prompt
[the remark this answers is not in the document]
zvone187: This is a leftover 😅 removed it

core/prompts/bug-hunter/iteration.prompt
```diff
@@ -0,0 +1,43 @@
...
+
+{% include "partials/user_feedback.prompt" %}
+
+{% if next_solution_to_try is not none %}
+Focus on solving this issue in the following way:
+```
+{{ next_solution_to_try }}
+```
+{% endif %}
+{#{% include "partials/doc_snippets.prompt" %}#}
```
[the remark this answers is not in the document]
zvone187: done

core/prompts/bug-hunter/iteration.prompt
```diff
@@ -0,0 +1,43 @@
...
+
+Based on this information, you need to figure out where is the problem that the user described. You have 2 options - to tell me exactly where is the problem happening or to add more logs to better determine where is the problem. If you think we should add more logs around the code to better understand the problem, tell me code snippets in which we should add the logs. If you think you know where the issue is, don't add any new logs but explain what log print tell point you to the problem, what the problem is, what is the solution to this problem and how the solution will fix the problem. What is your answer?
+
+{#**IMPORTANT**
```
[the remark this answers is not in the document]
zvone187: done

core/prompts/bug-hunter/get_bug_reproduction_instructions.prompt
```diff
@@ -0,0 +1,31 @@
...
+
+{% include "partials/user_feedback.prompt" %}
+
+{% if next_solution_to_try is not none %}
+Focus on solving this issue in the following way:
+```
+{{ next_solution_to_try }}
+```
+{% endif %}
+{#{% include "partials/doc_snippets.prompt" %}#}
```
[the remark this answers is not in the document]
zvone187: done

core/prompts/bug-hunter/bug_found_or_add_logs.prompt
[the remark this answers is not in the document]
zvone187: I think this messes LLM up. I would leave it like this but if you insist, I'll put the reasoning in the schema

core/config/magic_words.py
zvone187: ❤️

core/agents/troubleshooter.py
zvone187: you're right, fixing

core/agents/orchestrator.py
zvone187: changed

core/agents/orchestrator.py
zvone187: I made every if separate and added a comment under each of them to explain why is it being triggered

core/agents/orchestrator.py
zvone187: done

core/agents/bug_hunter.py
zvone187: Hm, I thought this is using Enum. Check `HuntConclusionType`. What is written here is just a description of an enum field (at least that's what I think)

core/agents/bug_hunter.py
zvone187: whoops, this is a leftover - removed it

core/agents/developer.py
zvone187: 😢

core/agents/developer.py
zvone187: changed to status in (...)

core/agents/bug_hunter.py
zvone187: will do

core/agents/bug_hunter.py
zvone187: I added the variable but splitting was so that it's more readable. Let me know if you would prefer going with "or" instead

core/agents/bug_hunter.py
zvone187: yup

core/agents/bug_hunter.py
zvone187: yup, changed

PR #1058, opened 2024-07-25
Add feature to spec

matija-ilijas:
Adding new feature to project specification on user request


core/agents/spec_writer.py
```diff
@@ -75,6 +83,27 @@ async def run(self) -> AgentResponse:
         self.next_state.action = SPEC_STEP_NAME
         return AgentResponse.done(self)
 
+    async def update_spec(self) -> AgentResponse:
+        feature_description = self.current_state.current_iteration["description"]
```
LeonOstrez, line 87: I understand that in this case it doesn't make too much difference but I would prefer using  `self.current_state.current_iteration["user_feedback"]` for code readability/understanding

core/agents/spec_writer.py
```diff
@@ -75,6 +83,27 @@ async def run(self) -> AgentResponse:
...
+        await self.send_message(
+            f"Adding feature with the following description to project specification:\n\n{feature_description}"
+        )
+        llm = self.get_llm()
+        convo = AgentConvo(self).template("add_new_feature", feature_description=feature_description)
+        llm_response: str = await llm(convo, temperature=0, parser=StringParser())
+        updated_spec = llm_response.strip()
+
+        self.next_state.specification = self.current_state.specification.clone()
+        self.next_state.specification.description = updated_spec
```
LeonOstrez, line 97: we should always have original description, so maybe it is good to add new column in `specification` table e.g. `original_description`

core/agents/spec_writer.py
```diff
@@ -75,6 +83,27 @@ async def run(self) -> AgentResponse:
...
+
+        telemetry.set(
+            "initial_prompt", updated_spec
```
LeonOstrez, line 100: let's keep `initial_prompt` the way it is now, but add  `updated_prompt`

PR #1098, opened 2024-10-01
Update spec_writer.py

CarlosReyesPena:
close_diff doesn't exist. fixing import project.


CarlosReyesPena:
This would allow users to continue working with import.

PR #1158, opened 2025-08-31
Fix rate limiting sleep duration calculation

### LINKED ISSUE #1121: [Bug]: Rate limit Timeout Sleep WAY too long
### Version

Command-line (Python) version

### Operating System

Ubuntu Linux

### What happened?

When GPT Pilot runs into openai.RateLimitError, it will sleep for well over what is necessary based on the response of the API..

openai.RateLimitError: Error code: 429 - {'error': {'message': 'Rate limit reached for gpt-4o in {REMOVED FOR BUG REPORT} on tokens per min (TPM): Limit 30000, Used 29328, Requested 6310. **Please try again in 11.276s**. Visit https://platform.openai.com/account/rate-limits to learn more.', 'type': 'tokens', 'param': None, 'code': 'rate_limit_exceeded'}}

GPT Pilot should ideally sleep for the suggested time in the response of 11.276s (maybe add 1 second). Instead, it sleeps for either 7200 seconds or 9600 seconds.. sometimes it prints both. And I have let it sit for hours and it never seems to start up again.

I dug into the source code and I think I may be able to fix this, and if its necessary I will create a pull request to merge into the main repository, but I wanted to first make sure that I am not missing something and just need to change the config.json, or something similar. 

Thank you in advance, and if you have had this issue too and there is no obvious fix, please let me know and I will start to fix it myself and create a pull request. I'd love to contribute to this awesome project!

onkardahale:
## Problem
Users reported excessive sleep times (7200+ seconds) during rate limiting, making the tool unusable. The issue was caused by using `wait_time.seconds` instead of `wait_time.total_seconds()` in the sleep logic

## Solution
  - Replace `wait_time.seconds` with `wait_time.total_seconds()` in base sleep logic
  - Add 1-hour maximum sleep duration cap with validation bounds
  - Implement base class fallback for error message parsing when headers missing
  - Added a test case for this scenario
  
## Impact
Rate limiting no longer makes the tool unusable. Sleep durations now respect API suggestions with reasonable upper bounds. I tested the exact scenario from the original issue and confirmed the fix resolves the problem.

Fixes #1121


onkardahale:
@LeonOstrez

